PR #16, opened 2019-12-14
Account for Null Value in Bitrate

robejo9:
Added does not equal null as modifier to lower then default to avoid 0 bitrate being input as a value.


robejo9:
Wrong modifier / resubmitting

PR #21, opened 2019-12-22
Update to lower potential errors

robejo9:
-Added mitigation for embeded pictures to avoid errors during transcode.
-Cleaned up unnecessary code.


robejo9:
adding jpeg support

PR #27, opened 2020-01-15
 Fix ID on TheRealShadoh plugins

JackDallas:
It's been found that the `Copy id` button doesn't work for these plugins as `.js` is included in the id field

I opened this on https://github.com/TheRealShadoh/Tdarr_Plugins/pull/1 as well but I'm not sure if they're active

### DIFF
--- a/Community/Tdarr_Plugin_z0ab_TheRealShadoh_FFmpeg_Subs_H264_Medium.js
+++ b/Community/Tdarr_Plugin_z0ab_TheRealShadoh_FFmpeg_Subs_H264_Medium.js
@@ -4,7 +4,7 @@
 function details() {
 
   return {
-    id: "Tdarr_Plugin_z0ab_TheRealShadoh_FFmpeg_Subs_H264_Medium.js",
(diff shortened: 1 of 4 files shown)


HaveAGitGat:
Hi Jack thanks for the PR.

All plugins should normally have the ‘requeueAfter’ switch set to true. Users may be using other plugins below the Shado plugins in their plugin stack which won’t work if this PR is merged. 

If endless requeueing is happening with the Shado plugins then it indicates that the plugins are not performing checks/changes correctly.

I’d prefer the plugin logic to be fixed rather than slapping a quick fix on by changing the requeueAfter setting. 

You can copy the plugins into your local plugins folder and apply your changes if you like as a quick fix.

JackDallas:
Ahhh okay I see, in that case I'll investigate further and work out what's causing this endless re-queue, and update this, thanks @HaveAGitGat

JackDallas:
Actually @HaveAGitGat would you accept this as just a fix for the name's and I'll open another PR when I solve the re-queue issue as the copy id button is broke on all these plugins as a result ?

HaveAGitGat:
Thanks. Sure will do 👍

JackDallas:
Thanks man :D

PR #39, opened 2020-02-06
Patch

controlol:
Fixes some variables from details() 
New plugin Tdarr_Plugin_e5c3_CnT_Remove_Letterbox.js


HaveAGitGat:
Change the `execSync(` to use `otherArguments.ffmpegPath` so it works cross platform.

PR #42, opened 2020-02-09
Migz Plugin: Bitrate Transcoding

Migz93:
Change plugin to work via bitrate related targets instead of filesize related target.


Migz93:
Don't accept this just yet, just found a bug

PR #45, opened 2020-02-12
Add Generic FFmpeg preset plugin h264/mp4

JackDallas:
This is very similar to theRealShadoh's plugins but it merges them all in to one and give the user the option to set the preset from the inputs and handles cases when the file has incompatible subs by dropping them


HaveAGitGat:
Thanks!

PR #46, opened 2020-02-13
WIP: Add linting & ci

JackDallas:
Okay this changes a lot but gives the repo linting, I'm super open to feedback on the rules and will happily keep changing and rebasing this until everyone is happy, the eslint also picked up a few instances that may have lead to bugs so I personally would say this is well worth having :)


HaveAGitGat:
Added some new rules which are being applied to new additions/edits, thanks though :)

PR #49, opened 2020-02-25
HQ Anime/Hand Drawn Animation Plugin

miasmaejuices:
Specifically tuned for use with Animation, primarily Anime with English subtitles.  Preserves Subtitles; Metadata; Attachments; Fonts and Styling.  CRF 18 to keep as much quality as reasonable, while keeping file sizes low.  Audio converted to AAC 512K for compatibility.  Have used `x265-params`, as it seems the source script I used failed to do so, and the variables were not being read properly.  Medium preset, need a reasonably fast CPU/Server.  10-Bit video output to help curb banding.


HaveAGitGat:
Nice!

PR #61, opened 2020-04-02
Subtitle Fix

miasmaejuices:
Added fix for subtitles per https://discordapp.com/channels/623392507828371476/626268797472604160/695001638628360212

Fixes error:
`
Stream map '0:s' matches no streams.
To ignore this, add a trailing '?' to the map.
`


HaveAGitGat:
Thanks!

PR #66, opened 2020-04-11
Added community plugin to set a default audio stream

SpencerReiter:


HaveAGitGat:
Thanks!

Mushin:
Would it be possible to add the option for multiple ranked preferred audio channels? I.e. 7.1, then 6.1, then 5.1?

PR #67, opened 2020-04-12
Migz2CleanTitle: Map all streams to output

rbraunschweig:


Migz93:
@HaveAGitGat happy for you to merge this, without the -map 0 that I've forgotten to put in it means only the first stream is copied i think.

rbraunschweig:
Well, the first stream of each codec type. I think you missed it because `-map 0` is placed before the ffmpeg command in your other plugins.

HaveAGitGat:
👍

PR #74, opened 2020-05-22
Add Tdarr_Plugin_drdd_standardise_all_in_one

dambler:
New plugin that does a bunch of things in a single pass rather than multiple (as would be required using mutliple plugins).

- Converts video to h265 where possible (there is a bitrate cut off).
- Ensures in MKV.
- Removes unwanted subtitle languages and commentary subtitles.
- Ensures multi channel audio is AC3 encoded.
- Removes commentary audio tracks.
- Supports both Intel Quick Sync and Nvidia NVENC encoding.

Edit: Should really say, a lot of the grunt work for this plugin was done by Migz - I used a lot of his logic and ideas!

### DIFF
--- a/Community/Tdarr_Plugin_drdd_standardise_all_in_one.js
+++ b/Community/Tdarr_Plugin_drdd_standardise_all_in_one.js
@@ -1,0 +1,417 @@
+function details() {
+  return {
+    id: "Tdarr_Plugin_drdd_standardise_all_in_one",
+    Name: "DrDD H265 MKV AC3 audio subtitles [QSV & NVENC]",
+    Stage: "Pre-processing",
+    Type: "Video",
+    Operation: "Transcode",
+    Description:
+      "In a single pass ensures all files are in MKV containers and where possible encoded in h265 (settings dependant on file bitrate), converts all multi channel audio to AC3, removes audio commentary and removes subtitles that are not in the configured language or marked as commentary. This plugin is opinionated based on how I like my library to be configured and based on the work done by Migz with his plugins (Thanks!).",
+    Version: "1.0",
+    Tags: "pre-processing,ffmpeg,qsv h265, nvenc h265",
+    Inputs: [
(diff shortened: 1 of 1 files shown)


HaveAGitGat:
TY

josh-gaby:
@dambler the last update to this was called "QSV Mpeg4 fix" but it also forces mpeg4 files to be software encoded when using NVENC, was that done intentionally due to some known issue or should it be safe to re-enable HW transcoding for mpeg4 if using NVENC?

milindpatel63:
@dambler @josh-gaby I too have similar question....my mpeg4 videos are not using gpu for transcoding....can we remove that "if statement" and re-enable hw transcoding for mpeg4?
i checked my gpu supported codecs using "ffmpeg -codecs", and it shows mpeg4_cuvid is supported...

milindpatel63:
@josh-gaby thankss...will do the same...
I'll just remove this or statement..
if (
        (inputs.qsv !== "true" && inputs.nvenc !== "true") _**||
        stream.codec_name === "mpeg4"**_
      ) 

and i'll add another line..
else if (file.video_codec_name == "mpeg4") {
            configuration.AddInputSetting("-c:v mpeg4_cuvid");
          }

Is this correct?
Thanks for the quick reply though :D

josh-gaby:
I changed the if statement as follows:
```
if (
        (inputs.qsv !== "true" && inputs.nvenc !== "true") ||
        stream.codec_name === "mpeg4"
) {
```
becomes
```
if (
        (inputs.qsv !== "true" && inputs.nvenc !== "true") || 
        (inputs.qsv == "true" && stream.codec_name === "mpeg4")
) {
```

that way the QSV/mpeg4 fix remains in place.

milindpatel63:
@josh-gaby okk...thanks a lot (Y)

dambler:
Nice one!

PR #75, opened 2020-05-22
JarBinks set of 2 new plugins

JarBinks:
After playing with Tdarr for a while and using it as the basis to do what I wanted, these are what I came up with.

They are not for the feint of heart, nor are they easily consumable.  I just hope they are useful, even in  a small way,  to someone.  I know they are to me!

A detailed explanation is included in the top of each file.


HaveAGitGat:
TY

PR #76, opened 2020-05-23
Migz Plugins - Fixes

Migz93:
Change both ffmpeg plugins reQueueAfter to false. These plugins should not need to be requeued and worked on again.

Move mkv data stream check to after extraArguments var creation so that it works as intended.
Updated mkv data check from "" to `` for consistency.

Correct typo in CleanSubs plugin. Thanks zzeneg for spotting this.


HaveAGitGat:
TY

PR #78, opened 2020-05-27
Add plugin to transcode video to H.265 HEVC using Intel QuickSync CPUs

Mthrboard:
This plugin transcodes video streams to H.265 HEVC using Intel QuickSync for hardware acceleration. It uses ffmpeg instead of HandBrake because ffmpeg can use hardware acceleration for decode and encode.


HaveAGitGat:
Thanks

PR #83, opened 2020-06-15
Update Tdarr_Plugin_MC93_Migz1FFMPEG_CPU.js

jayrox:
if `inputs` isn't defined, then `inputs.container` causes a plugin error that is not caught by the `if (inputs.container == "")`

this change checks the `inputs` first and if not defined, allows the plugin to be skipped properly.

### DIFF
--- a/Community/Tdarr_Plugin_MC93_Migz1FFMPEG_CPU.js
+++ b/Community/Tdarr_Plugin_MC93_Migz1FFMPEG_CPU.js
@@ -65,7 +65,7 @@ function plugin(file, librarySettings, inputs) {
   };
 
   // Check if inputs.container has been configured. If it hasn't then exit plugin.
-  if (inputs.container == "") {
+  if (!inputs || inputs.container == "") {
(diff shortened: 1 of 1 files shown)


jayrox:
looking at other plugins, this will likely need to be done to them as well.

dambler:
`inputs` is a method argument variable though, defined on line 57. What scenario are you running into where inputs is undefined entirely?

jayrox:
Fresh installed yesterday, created a test library with a bunch of .mkv files, added them to the queue and each ended up in the error queue.

Pulled up the docker logs and each one had a javascript error on that line saying it could not reference property of undefined. I edited the code locally to console.log the inputs param to see what it said, which returned undefined.

The reason it was undefined was that I had not gone into the Libraries > Library > Transcode options and set the container property. (BTW, This setting page is pretty hard to find as a newbie.)

HaveAGitGat:
Sorry missed this. Thanks.

PR #89, opened 2020-07-13
Adding Community Plugin

mdoom23:
Adding a new plugin inspired by other existing ones.  Configurable, Tiered, NVENC, and audio and subtitle clean-up all-in-one.

### DIFF
--- a/Community/Tdarr_Plugin_DOOM_NVENC_Tiered_MKV_CleanAll.js
+++ b/Community/Tdarr_Plugin_DOOM_NVENC_Tiered_MKV_CleanAll.js
@@ -1,0 +1,447 @@
+function details() {
+  return {
+    id: "Tdarr_Plugin_DOOM_NVENC_Tiered_MKV_CleanAll",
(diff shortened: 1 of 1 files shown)


HaveAGitGat:
Thanks. The only issue is ffprobe sometimes (or often for some people) doesn't provide the video stream bitrate and using the overall bitrate can lead to all sorts of expected results. I've added mediaInfo to the file scanner for the next release which is more reliable at getting the stream bitrates so it might be worth waiting until then to add this in.

HaveAGitGat:
Also `reQueueAfter` should be set to `true` in case there are other plugins below this plugin in the plugin stack.

mdoom23:
Thanks for checking things out!  Yeah, I definitely have stumbled on that issue with ffprobe a lot between my own plugins and other peoples.  I do look forward to having mediaInfo available.
I've been 'okay' with using overall bitrate because the only down side for me has been i end up not compressing as much as I could, but still will reduce file size.  And Good catch on reQueueAfter.

PR #98, opened 2020-09-04
Added 2 pass Loudnorm audio normalisation plugin, added file permissions plugin

jakejones48:
Normalization Plugin:
Plugin uses a two pass approach to loud norm volume normalization. Achieved by writing the output of ffmpeg to a temp file in the source directory alongside the video file. This is read during the second pass to get the required values. Temp file is removed after second pass. MKV is required to allow the file to have custom metadata so the plugin knows what stage it is up to.

Issues addressed in main Tdarr:
https://github.com/HaveAGitGat/Tdarr/issues/272
https://github.com/HaveAGitGat/Tdarr/issues/127

File Permission plugin:
Uses chown nobody:users to set file permissions for unraid.

helps with issue:
https://github.com/HaveAGitGat/Tdarr/issues/170
https://github.com/HaveAGitGat/Tdarr/issues/134


HaveAGitGat:
Thanks :)

PR #110, opened 2020-10-30
Add config option for audio transcode bitrate

tmchow:
Give option to specify the bitrate of the transcoded audio


HaveAGitGat:
Hi thanks for the contribution.

The reason I haven't merged this is because, for people already using this plugin, all new files will be moved into the transcode error/cancelled tab due to no bitrate being entered (Tdarr pulls community plugin updates hourly). When modifying existing plugins it's best to do it in such a way that existing users won't be affected 👍.

tmchow:
@HaveAGitGat good call. Let me refactor it a bit to make it optional.

tmchow:
@HaveAGitGat Just adjusted to skip it if not provided.

HaveAGitGat:
Thanks :)

PR #113, opened 2020-12-05
Migz Plugins

Migz93:
1) Fix Flawed logic in FFMPEG plugins which would cause remux conditions to never trigger. Was checking if file was HEVC & VP9 which is impossible as they're both codecs.

2) Fixed a bug with CleanTitle where plugin would fail if a stream in the file didn't have a title.

3) Include new plugin to just perform remuxes. Mainly aimed at remuxing to mkv or mp4.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,9 +2,9 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ master ]
+    branches: ['**']
   pull_request:
-    branches: [ master ]
+    branches: ['**']
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


HaveAGitGat:
Hi thanks for the PR. In order to increase code quality and catch bugs, all new plugin edits/additions will need to pass the lint checks. Most of the errors can be fixed by cloning the branch and running:

```
npm i
npm run lint:fix
```

There will still be errors which will need to be corrected manually. It would be appreciated if you could have a go at this else I'll try and do it when I get time (gotta work my way through a backlog of 68 plugins :( ).
Thanks

PR #117, opened 2020-12-15
Migz Plugins Update

Migz93:
1. Made changes to all plugins following lint testing. There are still various warnings/errors but my very limited knowledge of javascript means I just don't understand what the errors mean. Or in some cases I don't understand how to split the variable across several lines to conform to the line length limits.
I did manage to get it from:
952 problems (943 errors, 9 warnings)
down to:
37 problems (28 errors, 9 warnings)

2. Fix Flawed logic in FFMPEG plugins which would cause remux conditions to never trigger. Was checking if file was HEVC & VP9 which is impossible as they're both codecs.

3. Fixed a bug with CleanTitle where plugin would fail if a stream in the file didn't have a title.

4. Modify CleanSubtitle plugin to not remove sdh subtitles.

5. Include new plugin to just perform remuxes. Mainly aimed at remuxing to mkv or mp4.

### DIFF
--- a/Community/Tdarr_Plugin_MC93_Migz1FFMPEG.js
+++ b/Community/Tdarr_Plugin_MC93_Migz1FFMPEG.js
@@ -1,253 +1,286 @@
-/* eslint-disable */
+/* eslint no-plusplus: ["error", { "allowForLoopAfterthoughts": true }] */
 function details() {
   return {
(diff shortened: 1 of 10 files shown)


HaveAGitGat:
Looks good :)

PR #119, opened 2020-12-22
Tdarr_Plugin_rr01_drpeppershaker_extract_subs_to_SRT

drpeppershaker:
new plugin to extract SRT subtitles using Tdarr without calling exec() and uses only one command to extract--which allows the progress bar to work, and is faster than other available plugins as it does not need to re-queue.


HaveAGitGat:
Looks good, thanks!

PR #120, opened 2021-01-09
Migz Plugins Update

Migz93:
Just a small description field only update, no changes to how the plugins themselves actually work.

1) Update CPU plugin description as it mentions using a GPU which is incorrect.

2) Update CPU & GPU plugins to remove `\n` and `\\n` from the description fields as the description field doesn't seem to use these as new lines and were instead displaying as text.


HaveAGitGat:
Ty

PR #135, opened 2021-03-04
Adding two 265 plugins

weslocke:
Easily configurable plugins to use QSV and NVENC to convert directly to 265 in an MKV container, with options for scaling, bitrate adjustment, target quality adjustment and encoding profile used by engine.  Plugins are separated so that QSV and NVENC can be run simultaneously on the same machine


weslocke:
Quick fix for allowing presets to be set with numeric value as well as the keyword

HaveAGitGat:
Hello, thanks for opening a PR. Please don't disable eslint in the plugins. So this line needs to be removed from the top of each file:

`/* eslint-disable */`

PR #139, opened 2021-03-10
Create Tdarr_Plugin_ER01_Transcode audio and video with HW (PC and Mac)

erichter2018:
Files not in H265 will be transcoded into H265 using hw with ffmpeg, assuming mkv container. Plugin uses QS if the node runs on a PC, or Videotoolbox if run on a Mac.  Much thanks to Migz for bulk of the important code.  Quality is controlled via bitrate adjustments - H264 to H265 assumes 0.5x bitrate.  Resolution change from 1080p to 720p assumes 0.7x bitrate.  Audio conversion is either 2 channel ac3 or 6 channel ac3, for maximal compatibility and small file size.  All subtitles removed.  The idea is to homogenize your collection to 1080p or higher movies with 5.1 audio, or 720p TV shows with 2.0 audio.`,


HaveAGitGat:
Thanks

PR #147, opened 2021-04-19
Updating Plugin to use file object

zachg99:
I've updated these plugins to use the file object provided by Tdarr rather than running mediainfo and ffprobe on their own. I've also updated the comments section to remove the requirement to install extra packages and included updated Tdarr v2.0x docker compose file commands.


HaveAGitGat:
Thanks 👍

PR #153, opened 2021-05-17
Fixed a typo causing NaN bitrates

zachg99:
Fixed a typo (missing space) causing the script to not read the video and audio bitrate for some files resulting in a larger file after processing.

### DIFF
--- a/Community/Tdarr_Plugin_JB69_JBHEVCQSV_MinimalFile.js
+++ b/Community/Tdarr_Plugin_JB69_JBHEVCQSV_MinimalFile.js
@@ -761,7 +761,7 @@ function findMediaInfoItem(file, index) {
 			currMIOrder = file.mediaInfo.track[i].ID - 1;
 		}
 
-		if (currMIOrder == index|| currMIOrder == "0-" + index) {
+		if (currMIOrder == index || currMIOrder == "0-" + index) {
 			return i;
 		}
 	}


zachg99:
Hi, can this fix be merged?

zachg99:
Thanks!

I thought so too but it wasn’t evaluating true without the space for files that fell into that path.

--
Our problems are like diamonds. We inflate their value because we don't understand how many they have in Africa.

tehniemer:
I'm still seeing some files generate NaN bitrates and creating larger files after processing, I'm working to narrow down the codecs or containers that cause the issue, but I wanted to let you know that it's not fixed.

zachg99:
Oh thanks! That's annoying. I hadn't had the issue after I added that
space, even using the files that previously failed. It didn't seem to be
processing the OR operator without the space so it wasn't finding a
matching stream, or something like that.

If the script decides the file has a NaN bitrate it then uses a default
bitrate that's rather arbitrary. If you can figure it out that would be
awesome.

--

Our problems are like diamonds. We inflate their value because we don't
understand how many they have in Africa.

tehniemer:
Where in the script is the arbitrary bitrate set? It's not obvious to me. I've used your script as a base and made some modifications to it to suit my needs and I think using the original stream bitrate vs something arbitrary would work better for me in this case.

zachg99:
It's been a while, but line 509 is where it calculates the optimal video
bitrate based on the height, width, and fps of the video file. At this
point the actual bitrate might be lower (or higher) in the file than what
the script calculated as optimal so the next step that matters in this
process is like 542. The videoBR is NaN so that if statement will evaluate
as false dropping to the else statement. So, it skipped lines 554-558 that
would have adjusted the target bitrate so the file won't increase in size.

Side note, in the else block boltranscodeVideo = true is commented out but
it was initialized to true so it's going to try to transcode.

If your video file had a lower bitrate than the calculated optimal, happens
a lot, when it transcodes the file it will also attempt to increase the
bitrate and thereby increase the file size.

--

Our problems are like diamonds. We inflate their value because we don't
understand how many they have in Africa.

tehniemer:
OK, I think I get it now. That's the part of the code I pretty significantly changed for my needs, I now have a catch for the NaN bitrate to transcode to a user defined minimum. Here's the code if you care to take a look.
```	//We need to set the minimum bitrate
	if ((videoheight * videowidth) > minvideopixels4K) {
	    response.infoLog += "Video stream determined to be 4K. Minimum bitrate set as: " + (minbitrate4K/1000) + " kbps. \n";
	    minimumvideobitrate = minbitrate4K;
	} else if ((videoheight * videowidth) > minvideopixels1080p) {
	    response.infoLog += "Video stream determined to be 1080p. Minimum bitrate set as: " + (minbitrate1080p/1000) + " kbps. \n";
	    minimumvideobitrate = minbitrate1080p;
	} else if ((videoheight * videowidth) > minvideopixels720p) {
	    response.infoLog += "Video stream determined to be 720p. Minimum bitrate set as: " + (minbitrate720p/1000) + " kbps. \n";
	    minimumvideobitrate = minbitrate720p;
	} else {
	    response.infoLog += "Video stream determined to be 480p or lower. Minimum bitrate set as: " + (minbitrate480p/1000) + " kbps. \n";
	    minimumvideobitrate = minbitrate480p;
	}

	//Now calculate the optimal bitrate 
	optimalvideobitrate = Math.floor((videoheight * videowidth * targetframerate) * targetcodeccompression);

	if (videoBR < optimalvideobitrate) {
		//We need to be careful here or else we could produce a bad quality stream
		response.infoLog += "Low source bitrate! \n";	
		if (file.ffProbeData.streams[videoIdx].codec_name == targetvideocodec) {
			response.infoLog += "Codec is already " + targetvideocodec + " and \n";
			if (bolSource10bit == boluse10bit) {
	            response.infoLog += "source stream bitrate, " + Math.floor(videoBR/1000) + "kbps, is less than optimal bitrate, " + (optimalvideobitrate/1000) + "kbps. Using source stream. \n";
	            optimalvideobitrate = videoBR;
				boltranscodeVideo = false;
			} else {
	            response.infoLog += "source stream bit depth is different from target. Transcoding with source bitrate. \n";
				optimalvideobitrate = videoBR;
			}
		} else {
			//We have a codec change with not much meat so we need to adjust our target bitrate
			response.infoLog += "Source bitrate, " + Math.floor(videoBR/1000) + "kbps, is less than optimal bitrate, " + (optimalvideobitrate/1000) + " kbps. \n";
			if (Math.floor(videoBR * targetreductionforcodecswitchonly) > minimumvideobitrate) {
				response.infoLog += "Transcoding with a codec change using " + Math.floor(targetreductionforcodecswitchonly * 100) + "% of source bitrate. \n";
				optimalvideobitrate = Math.floor(videoBR * targetreductionforcodecswitchonly);
			} else if (videoBR < minimumvideobitrate) {
				response.infoLog += "Transcoding with a codec change using source bitrate. \n";
				optimalvideobitrate = videoBR;
			} else {
				response.infoLog += "Transcoding with a codec change using minimum acceptable bitrate. \n";
				optimalvideobitrate = minimumvideobitrate;
			}
		}
	} else if (isNaN(videoBR)) {
		//Cannot determine source bitrate
		response.infoLog += "Cannot determine source bitrate, throwing in towel and using minimum acceptable bitrate. \n";
		optimalvideobitrate = minimumvideobitrate;
	} else {
		//Source bitrate has enough meat for a decent transcode
		response.infoLog += "Source bitrate, " + Math.floor(videoBR/1000) + "kbps, is high enough to transcode to optimal bitrate. \n";
	}
	response.infoLog += "Post-process video stream will be: " + videoheight + " x " + videowidth + ", " + videoFPS + ", " + Math.floor(optimalvideobitrate/1000) + " kbps \n";
```

zachg99:
Nice, that's a good fix. What did you set the minimumvideobitrate to for
each of the stream types?

--

Our problems are like diamonds. We inflate their value because we don't
understand how many they have in Africa.

tehniemer:
I created user configurable inputs for those, I have them set at 20,000, 5000, 2200, and 750kbps respectively. I spent a bunch of time working on the script this weekend so this section has changed yet again, but this method stays the same.

zachg99:
FYI I think i found the NaN bitrate issue and resolved it in the latest checkin. There were two things causing the problem, one was bad file tags, the other was some issues with the function that found the mediainfo track tied to the ffprobe stream.

The main fix for the bad tags is to run:

	mkvpropedit --add-track-statistics-tags “v<video file>”

This requires installing the mkvtoolnix to the container. I’ve requested that this be added, until then you can install it manually via: apt install mkvtoolnix (from within the container).

The issue with the funtion is that it wasn’t handling the General mediainfo “track” properly. Now it should and return the proper track for processing.

Let me know if you see any issues with the latest version (once/if it’s approved).

Thanks,
Zach

--
Our problems are like diamonds. We inflate their value because we don't understand how many they have in Africa.

Issue #156, opened 2021-06-01
Bug: Tdarr_Plugin_b39x_the1poet_surround_sound_to_ac3 converts 7.1 audio to 5.1


17hoehbr:
Hi, I’ve been using this plugin with my setup for a few days now and it seems to doubts a great job converting my AAC 5.1 tracks to AC3. However today I noticed that when I ran a video that had a 7.1 audio track it down mixed to 5.1. 

Attached are a few screenshots from the transcode info that show the audio streams before and after going through my plugin stack. I also run Tdarr_Plugin_MC93_Migz5ConvertAudio right before transcoding the surround sound which is why the output file has a stereo AAC track and a duplicate 5.1 track.
![8A2E3BD1-245A-42DC-BD9F-B21DAC0B0B49](https://user-images.githubusercontent.com/1351017/120292314-34b19880-c292-11eb-91d1-aceef6bd019a.jpeg)
![8FFAA44D-BEED-472B-9BFA-7B0EDC16E527](https://user-images.githubusercontent.com/1351017/120292318-35e2c580-c292-11eb-8635-39257b8aaa1c.jpeg)
![F271CED9-3104-4514-A35C-0EEC805BD13E](https://user-images.githubusercontent.com/1351017/120292320-367b5c00-c292-11eb-8ddd-04cf289d115a.jpeg)

(Apologies for the small screenshots since I’m on my phone right now but I can hop on my PC to add logs and other info if needed)

drummerdude152002:
Feel free to close this @17hoehbr . AC3 has a maximum channel count of 5 full-bandwidth channels, thus why you're seeing the downmix from 7.1 to 5.1. You can read more about that [here](https://en.wikipedia.org/wiki/Dolby_Digital_Plus)

PR #157, opened 2021-06-01
#156

17hoehbr:
Since FFMPEG does not support converting 7.1 to AC3 the plugin was previously downmixing 7.1 audio to 5.1. Since I'm also running Tdarr_Plugin_MC93_Migz5ConvertAudio to downmix tracks anyway, this caused me to end up with duplicate 5.1 tracks in all of my videos. This update does not fix the fundamental issue of ffmpeg not converting 7.1 to AC3, instead it just makes the plugin ignore audio streams with >6 channels since ffmpeg can't convert them to AC3 anyway. Combined with Tdarr_Plugin_MC93_Migz5ConvertAudio, this should allow users to have stereo and 5.1 audio streams encoded in AC3 while keeping the original 7.1 stream.

### DIFF
--- a/Community/Tdarr_Plugin_b39x_the1poet_surround_sound_to_ac3.js
+++ b/Community/Tdarr_Plugin_b39x_the1poet_surround_sound_to_ac3.js
@@ -48,7 +48,7 @@ function plugin(file) {
 
       try {
         if (
-          file.ffProbeData.streams[i].channels >= 6 &&
+          file.ffProbeData.streams[i].channels == 6 &&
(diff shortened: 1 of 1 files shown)


AbbieDoobie:
Can the description for this plugin please be updated? I just let tdarr go at my library for a week, and all the 7.1 streams don't have 5.1 AC3 equivalents which I need. The plugin's description mentions "surround sound", and it took hours of digging to figure out what went wrong. I'll need to figure something else out I guess.

Could including 7.1 > 5.1 downmixing just be an option?

PR #165, opened 2021-07-17
Create Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js

vodkapmp:
Same as my GPU based one basically, just for CPU

### DIFF
--- a/Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
+++ b/Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
@@ -8,7 +8,6 @@ function details() {
     Description: `[Contains built-in filter] This plugin uses different CRF values depending on resolution, 
        the CRF value is configurable per resolution.
        FFmpeg Preset can be configured, uses slow by default. 
-       ALL OPTIONS MUST BE CONFIGURED UNLESS MARKED OPTIONAL!
        If files are not in hevc they will be transcoded. 
        The output container is mkv. \n\n`,
     Version: '1.00',
@@ -20,43 +19,42 @@ function details() {
     Inputs: [
       {
         name: 'sdCRF',
-        tooltip: `Enter the CRF value you want for 480p and 576p content. (0-51, lower = higher quality, bigger file)
+        tooltip: `Enter the CRF value you want for 480p and 576p content. Defaults to 20 (0-51, lower = higher quality, bigger file)
          \\nExample:\\n 
         
         19`,
       },
       {
         name: 'hdCRF',
-        tooltip: `Enter the CRF value you want for 720p content. (0-51, lower = higher quality, bigger file)
+        tooltip: `Enter the CRF value you want for 720p content. Defaults to 22 (0-51, lower = higher quality, bigger file)
         
         \\nExample:\\n
         21`,
       },
       {
         name: 'fullhdCRF',
-        tooltip: `Enter the CRF value you want for 1080p content. (0-51, lower = higher quality, bigger file)
+        tooltip: `Enter the CRF value you want for 1080p content. Defaults to 24 (0-51, lower = higher quality, bigger file)
         
         \\nExample:\\n
         23`,
       },
       {
         name: 'uhdCRF',
-        tooltip: `Enter the CRF value you want for 4K/UHD/2160p content. (0-51, lower = higher quality, bigger file)
+        tooltip: `Enter the CRF value you want for 4K/UHD/2160p content. Defaults to 28 (0-51, lower = higher quality, bigger file)
         
         \\nExample:\\n
         26`,
       },
       {
         name: 'bframe',
-        tooltip: `Specify amount of b-frames to use, 0-16.
+        tooltip: `Specify amount of b-frames to use, 0-16, defaults to 8.
         
         \\nExample:\\n
         8`,
       },
       {
         name: 'ffmpegPreset',
-        tooltip: `OPTIONAL, DEFAULTS TO SLOW IF NOT SET 
-        \\n Enter the ffmpeg preset you want, leave blank for default (slow) 
+        tooltip: `Enter the ffmpeg preset you want, leave blank for default (slow) 
         
         \\nExample:\\n 
           slow  
@@ -108,7 +106,14 @@ function plugin(file, librarySettings, inputs) {
   // if we made it to this point it is safe to assume there is no hevc stream
   response.infoLog += '☒File is not hevc!\n';
 
-  // Check if preset is configured, default to slow if not
+  // set sane input defaults if not configured
+  const sdCRF = inputs.sdCRF ? inputs.sdCRF : 20;
+  const hdCRF = inputs.hdCRF ? inputs.hdCRF : 22;
+  const fullhdCRF = inputs.fullhdCRF ? inputs.fullhdCRF : 24;
+  const uhdCRF = inputs.uhdCRF ? inputs.uhdCRF : 28;
+  const bframe = inputs.bframe ? inputs.bframe : 8;
+
+  // set preset to slow if not configured
   let ffmpegPreset = 'slow';
   if (!inputs.ffmpegPreset) {
     response.infoLog += '☑Preset not set, defaulting to slow\n';
@@ -121,16 +126,16 @@ function plugin(file, librarySettings, inputs) {
   switch (file.video_resolution) {
     case '480p':
     case '576p':
-      crf = inputs.sdCRF;
+      crf = sdCRF;
       break;
     case '720p':
-      crf = inputs.hdCRF;
+      crf = hdCRF;
       break;
     case '1080p':
-      crf = inputs.fullhdCRF;
+      crf = fullhdCRF;
       break;
     case '4KUHD':
-      crf = inputs.uhdCRF;
+      crf = uhdCRF;
       break;
     default:
       response.infoLog += 'Could for some reason not detect resolution, plugin will not proceed. \n';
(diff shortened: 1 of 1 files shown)


HaveAGitGat:
New plugins shouldn’t have eslint disabled at the top 😁.

HaveAGitGat (approved):

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,201 @@
+/* eslint-disable */
```
HaveAGitGat, line 1: Remove this please 👍 : /* eslint-disable */

HaveAGitGat (changes requested):

vodkapmp:
I'll fix all this ~~pointless~~ linting sometime later today.

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,200 @@
...
+      response.preset += `,${map} -dn -c:v libx265 -preset ${ffmpeg_preset} -x265-params crf=${inputs.fullhdCRF}:bframes=${inputs.bframe}:rc-lookahead=32:ref=6:b-intra=1:aq-mode=3 -a53cc 0 -c:a copy ${subcli}${maxmux}`
+      transcode = 1
+    }
+    //file will be encoded if the resolution is 4K
+    //codec will be checked so it can be transcoded correctly
+    if (file.video_resolution === '4KUHD') {
+      CRFinuse = `${inputs.uhdCRF}`
+      response.preset += `,${map} -dn -c:v libx265 -preset ${ffmpeg_preset} -x265-params crf=${inputs.uhdCRF}:bframes=${inputs.bframe}:rc-lookahead=32:ref=6:b-intra=1:aq-mode=3 -a53cc 0 -c:a copy ${subcli}${maxmux}`
+      transcode = 1
+    }
```
supersnellehenk, line 185: Perhaps you could clean this up a bit by creating a switch that sets `CRFinuse`, then using that for a single `response.preset` instead of these 4 if statements. You could use the `default` case of the switch to abort the plugin.

LargoUsagi:
It may be useful to add an optional configuration to supply a maximum bitrate so you can ensure the files can meet network constraints.

vodkapmp:
> It may be useful to add an optional configuration to supply a maximum bitrate so you can ensure the files can meet network constraints.

Absolutely not, that defeats the whole point of using a crf based approach.
Use a two-pass approach (which tdarr does not really support) if you need an absolute maximum bitrate.

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,150 @@
...
+      break;
+    }
+  }
+
+  // if we made it to this point it is safe to assume there is no hevc stream
+  response.infoLog += '☒File is not hevc!\n';
+
+  // Check if preset is configured, default to slow if not
+  let ffmpegPreset = 'slow';
+  if (inputs.ffmpegPreset === undefined) {
```
supersnellehenk, line 113: Change to `if (!inputs.ffmpegPreset)`

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,150 @@
...
+  // encoding settings
+  response.preset += `,-map 0 -dn -c:v libx265 -preset ${ffmpegPreset}`
+  + ` -x265-params crf=${crf}:bframes=${inputs.bframe}:rc-lookahead=32:ref=6:b-intra=1:aq-mode=3`
+  + ' -a53cc 0 -c:a copy -c:s copy -max_muxing_queue_size 9999';
+  response.infoLog += `☑File is ${file.video_resolution}, using CRF value of ${crf}!\n`;
+  response.infoLog += 'File is being transcoded!\n';
+
+  return response;
+};
+module.exports.details = details;
```
supersnellehenk, line 150: Either do it for the plugin too, or move this to line 1 pls

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,150 @@
...
+        
+        \\nExample:\\n 
+          veryfast`,
+      },
+    ],
+  };
+}
+
+module.exports.plugin = function plugin(file, librarySettings, inputs) {
+  let crf = '';
```
supersnellehenk, line 78: This can be `let crf;`, no need to instantiate it as an empty string

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,150 @@
...
+
+  // check if the file is already hevc
+  // it will not be transcoded if true and the plugin will be stopped immediately
+  for (let i = 0; i < file.ffProbeData.streams.length; i += 1) {
+    if (file.ffProbeData.streams[i].codec_name.toLowerCase() === 'hevc') {
+      response.processFile = false;
+      response.infoLog += '☑File is already in hevc! \n';
+      break;
+    }
+  }
```
supersnellehenk, line 106: Do we need this to be a loop, if the file is a video it's safe to assume the first stream is a video stream.

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
supersnellehenk: Would be neat if you flip the if so it could just become `if (inputs.ffmpegPreset)`

supersnellehenk (changes requested):

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
vodkapmp: Better safe than sorry, this definitely stays

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
vodkapmp: willdo

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
vodkapmp: > Change to `if (!inputs.ffmpegPreset)`

lets go with this then

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
vodkapmp: > Either do it for the plugin too, or move this to line 1 pls

Done

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
vodkapmp: This has been done

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -0,0 +1,151 @@
...
+  response.infoLog += '☑File is a video! \n';
+
+  // check if the file is already hevc
+  // it will not be transcoded if true and the plugin will be stopped immediately
+  for (let i = 0; i < file.ffProbeData.streams.length; i += 1) {
+    if (file.ffProbeData.streams[i].codec_name.toLowerCase() === 'hevc') {
+      response.processFile = false;
+      response.infoLog += '☑File is already in hevc! \n';
+      break;
```
supersnellehenk, line 104: Shouldn't this just return the response immediately? There's no point to run the rest of the plugin, as it won't transcode.

vodkapmp: you are completely correct

LargoUsagi:
Adding sane defaults and making all inputs optional would help, as well as either clamping the bframes value to a max of 16 or adding some logic to set the look ahead to minimally 2x of bframes.

I think that would allow this to be the defacto default CPU 265 plugin for tdarr.

vodkapmp:
> Adding sane defaults and making all inputs optional would help, as well as either clamping the bframes value to a max of 16 or adding some logic to set the look ahead to minimally 2x of bframes.
> 
> I think that would allow this to be the defacto default CPU 265 plugin for tdarr.

I would love to add sane defaults, but my sane defaults for CRF is very different from other peoples sane defaults.

Should probably clamp the b frames, but I was honestly hoping for a requested feature of tdarr with radio buttons and value sliders to be implemented so I could use a value slider for the bframes and crf instead..

LargoUsagi:
I think CRF 20 across the board, 8bframe and slow speed would be a decent starting point.  Better than the 20 constant plugin that seems to do nothing with bframes, and maybe adding into the description on what each thing does.

You could also just punch in your thought on sane defaults and some improved descriptions on the parameters because ffmpegs docs are a pain to read and understand for the lay person, so having some plain English on what these numbers mean.  Then people can make thier own choices as they have to do today, and if they don't they take your informed opinion.

vodkapmp:
> I think CRF 20 across the board, 8bframe and slow speed would be a decent starting point. Better than the 20 constant plugin that seems to do nothing with bframes, and maybe adding into the description on what each thing does.
> 
> You could also just punch in your thought on sane defaults and some improved descriptions on the parameters because ffmpegs docs are a pain to read and understand for the lay person, so having some plain English on what these numbers mean. Then people can make thier own choices as they have to do today, and if they don't they take your informed opinion.

I have decided to set some "sane" defaults, which are for me way too high crf values, but probably about what people "expect" to have in terms of compression of the files when going to h265. 
And also 8 bframes.

supersnellehenk (approved):

HaveAGitGat (approved):

PR #192, opened 2021-10-30
New autoscan plugin to trigger manual scan with authentication support

mikertoad:


supersnellehenk:
@mikertoad could you fix the eslint errors from the jobs? :)

mikertoad:
Think I've fixed everything now - thanks

supersnellehenk:
I'm nowhere near an autoscan expert, so what does this plugin do differently compared to `Tdarr_Plugin_MC93_MigzPlex_Autoscan`?

mikertoad:
I believe the existing one is based on an old version and a number of people have issues with it.
I created this to use the newer setup which I also believe doesn't require a passkey.

supersnellehenk:
Could you perhaps add the link of the autoscan in the description so we can keep the two apart? The Migz plugin uses https://github.com/l3uddz/plex_autoscan which hasn't been updated since 2019.

Community/Tdarr_Plugin_TD01_TOAD_Plex_Autoscan.js
```diff
@@ -0,0 +1,91 @@
...
+  const request = require('request');
+  // Set up required variables.
+  const ADDRESS = inputs.autoscan_address;
+  const PORT = inputs.autoscan_port;
+  let file2 = '';
+
+  let filepath = '';
+  const response = {};
+  filepath = `${file.file}`;
+  file2 = `${filepath.split('/').slice(0, -1).join('/')}/`;
```
supersnellehenk, line 54: Couldn't you just combine `filepath` and `file2` into a single variable? `filepath` isn't needed at all.

Community/Tdarr_Plugin_TD01_TOAD_Plex_Autoscan.js
```diff
@@ -0,0 +1,91 @@
...
+    id: 'Tdarr_Plugin_TD01_TOAD_Plex_Autoscan',
+    Stage: 'Post-processing',
+    Name: 'Trigger Plex_Autoscan.',
+    Type: 'Video',
+    Operation: '',
+    Description: `Connects to plex_autoscan and triggers a manual search within the files directory.\n\n
+    Works with the hotio autoscan docker that can be found here https://hotio.dev/containers/autoscan/ which \n\n
+    is based on https://github.com/cloudbox/autoscan`,
+    Version: '1.0',
+    Link: '',
```
supersnellehenk, line 17: Please enter the link to the file in the community folder.

Community/Tdarr_Plugin_TD01_TOAD_Plex_Autoscan.js
```diff
@@ -0,0 +1,91 @@
...
+  let filepath = '';
+  const response = {};
+  filepath = `${file.file}`;
+  file2 = `${filepath.split('/').slice(0, -1).join('/')}/`;
+
+  // Check if all inputs have been configured. If they haven't then exit plugin.
+  if (
+    inputs
+    && inputs.autoscan_address === ''
+    && inputs.autoscan_port === ''
```
supersnellehenk, line 60: You can do `!inputs.autoscan_address || !inputs.autoscan_port`. That way it'll fail when either of them is undefined or empty. Currently it'll only fail when both are empty.

supersnellehenk (changes requested):

Community/Tdarr_Plugin_TD01_TOAD_Plex_Autoscan.js
mikertoad: completed

Community/Tdarr_Plugin_TD01_TOAD_Plex_Autoscan.js
mikertoad: completed

supersnellehenk (approved):

deathbybandaid:
Plex_autoscan requires direct access to plex and it's database. Also only works for Plex.

Cloudbox/Autoscan uses the plex api, and also works for emby and jellyfin. Also, you can trigger s scan on multiple plex servers.

For people like me that mount storage over CIFS, there's no inotify events to trigger partial scans, so these scripts are a must have.

deathbybandaid:
After looking at this PR, references to "plex_autoscan" should be changed to "autoscan".

Also, Autoscan can be setup to require a username And password.

A python script I was working on would look like

````

url_headers = {'Authorization': 'Basic %s' % self.password}

scanpost = self.plugin_utils.web.session.post(
                "%s/triggers/manual?dir=%s" % (self.address_full, urllib.parse.quote(filepath)),
                headers=url_headers,
                auth=(self.username, self.password)
                )
````

**Comment by vodkapmp** [comment]
Since this pull request is for two plugins now, could you change the title of the pull request to reflect this?

**Comment by mikertoad** [comment]
Corrected id and quote format

File: Community/Tdarr_Plugin_TD01_TOAD_Autoscan.js [hunk 4]
```diff
@@ -0,0 +1,115 @@
...
+
+module.exports.plugin = function plugin(file, librarySettings, inputs) {
+  // eslint-disable-next-line global-require,import/no-unresolved,import/no-extraneous-dependencies
+  const request = require('request');
+  // Set up required variables.
+  const ADDRESS = inputs.autoscan_address;
+  const PORT = inputs.autoscan_port;
+
+  if (inputs.autoscan_username) {
+    // eslint-disable-next-line no-unused-vars
```
supersnellehenk, line 68: Could you please remove this? Just disabling eslint isn't the way to solve this.

Community/Tdarr_Plugin_TD01_TOAD_Autoscan.js
```diff
@@ -0,0 +1,115 @@
...
+    response.infoLog += '☒Plugin options have not been configured, please configure options. Skipping this plugin.  \n';
+    response.processFile = false;
+    return response;
+  }
+
+  // Set content of request/post.
+  request.post({
+    headers: {
+      'content-type': 'application/json',
+      // eslint-disable-next-line no-undef
```
supersnellehenk, line 92: Could you please remove this? Just disabling eslint isn't the way to solve this.

Community/Tdarr_Plugin_TD01_TOAD_Autoscan.js
```diff
@@ -0,0 +1,115 @@
...
+      Authorization: auth,
+    },
+    url: `${ADDRESS}:${PORT}/triggers/manual?dir=${filepath}`,
+  },
+  // eslint-disable-next-line no-unused-vars
```
supersnellehenk, line 97: Could you please remove this? Just disabling eslint isn't the way to solve this.

Community/Tdarr_Plugin_TD01_TOAD_Autoscan.js
```diff
@@ -0,0 +1,115 @@
...
+  (error, res, body) => {
+    if (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
+    // eslint-disable-next-line no-console
+    // console.log(`statusCode: ${res.statusCode}`);
+    // eslint-disable-next-line no-console
+    // console.log(body);
```
supersnellehenk, line 106: Could you also remove all commented out blocks of code?

supersnellehenk (changes requested):

deathbybandaid:
@mikertoad I was just wondering if you had a few minutes to tweak the above requested changes?

Not trying to nag, but I've eager for this plugin to be available.

HaveAGitGat:
Fixed the issues and added some default values which are supported in 2.00.12

deathbybandaid:
Thanks @HaveAGitGat for the merge!

PR #194, opened 2021-11-04
New plugin which downmixes to stereo and applies DNC

jenskaalen:
A plugin which aims to downmix and apply DNC so the viewing experience is more pleasant for those with stereo systems. This also counteracts the whisperwhisperBANG effect.


Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+            response.infoLog = "Didn't find file to be in stereo... Downmixing and applying DRC!";
+            return response;
+        }
+
+    }
+
+    return response;
+};
+
+module.exports.onTranscodeSuccess = function onTranscodeSuccess(
```
supersnellehenk, line 57: You can remove this function entirely.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+    const response = {
+        file,
+        removeFromDB: false,
+        updateDB: false,
+    };
+
+    return response;
+};
+
+module.exports.onTranscodeError = function onTranscodeError(
```
supersnellehenk, line 75: You can also remove this one.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+        const stream = file.ffProbeData.streams[index];
+        if (stream.codec_type === "audio" && stream.channels && stream.channels > 3) {
+            console.log('stream.channels isXX ' + stream.channels);
+            response.preset = '-sn <io> -vcodec copy -scodec copy -acodec aac -af "dynaudnorm,pan=stereo|FL < 1.0*FL + 0.707*FC + 0.707*BL|FR < 1.0*FR + 0.707*FC + 0.707*BR"';
+            response.processFile = true;
+            response.infoLog = "Didn't find file to be in stereo... Downmixing and applying DRC!";
+            return response;
+        }
+
+    }
```
supersnellehenk, line 52: This will only work for a single audio stream in the file.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+        FFmpegMode: false,
+        reQueueAfter: true,
+        infoLog: '', 
+        file,
+        removeFromDB: false, 
+        updateDB: false, 
+    };
+
+    response.container = `.${file.container}`;
+    response.FFmpegMode = true;
```
supersnellehenk, line 40: You can set these values as the default in the response object.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+module.exports.dependencies = [
+    'import-fresh',
+];
+
+module.exports.details = function details() {
+    return {
+        id: 'Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC',
+        Stage: 'Pre-processing', 
+        Name: 'Downmix & Dynamic range compression',
+        Type: 'Video',
```
supersnellehenk, line 11: This is an audio plug-in :)

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+        Operation: 'Transcode',
+        Description: 'Downmixes surround to stereo AND applies dynamic range compression. Files already in stereo will be skipped \n\n',
+        Version: '1.00',
+        Link: 'https://github.com/HaveAGitGat/Tdarr_Plugin_aaaa_Pre_Proc_Example',
```
supersnellehenk, line 15: Make this link point to the actual file if it were in the repo.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
...
+        Tags: 'ffmpeg', 
+
+        Inputs: [
+            
+        ],
+    };
+};
+
+module.exports.plugin = function plugin(file, librarySettings, inputs) {
+    const importFresh = require('import-fresh');
```
supersnellehenk, line 25: You can remove this, you're not using it.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
```diff
@@ -0,0 +1,89 @@
+// List any npm dependencies which the plugin needs, they will be auto installed when the plugin runs:
+module.exports.dependencies = [
+    'import-fresh',
+];
+
```
supersnellehenk, line 5: You can remove this, you're not using it.

supersnellehenk (changes requested):

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
jenskaalen: Oh, that didn't cross my mind. Rarely have multi-audio files. Suggestion on how to deal with this?

supersnellehenk: You'd have to loop through all the audio tracks, then probably either keep the original codec it's in or say everything's going to be converted to aac. You'd also have to deal with channels probably.

jenskaalen:
Suggesting a narrowing of scope: Only processing files with a single, surround, audio track.  
Either that or dropping the plugin altogether. Tried fiddling with looping but did not like the added complexity, so I don't want to commit something error prone.

Community/Tdarr_Plugin_jeons001_Downmix_to_stereo_and_apply_DNC.js
jenskaalen: Suggesting a narrowing of scope: Only processing files with a single, surround, audio track.
Either that or dropping the plugin altogether. Tried fiddling with looping but did not like the added complexity, so I don't want to commit something error prone.

HaveAGitGat:
Looks good :)

PR #195, opened 2021-11-04
Use user defined preset in ffmpeg command

evoactivity:
The preset is verified but never used, this PR uses the user defined preset when converting the video


supersnellehenk (approved):

supersnellehenk:
I'm amazed this has been wrong since the 12th of February 2020...

HaveAGitGat (approved):

PR #199, opened 2021-11-17
Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js

Greg-2600:
Trying to contribute a frequently mentioned feature, built and tested on a clustered Tdarr 2.00.10 in Debian/Ubuntu.

### DIFF
--- a/Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
+++ b/Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
@@ -1,35 +1,39 @@
-module.exports.dependencies = ['import-fresh'];
-
 // eslint-disable-next-line import/no-extraneous-dependencies
 const importFresh = require('import-fresh');
+const loadDefaultValues = require('../methods/loadDefaultValues');
 
-function details() {
-  return {
-    id: 'Tdarr_Plugin_Greg_MP3_FFMPEG_CPU',
-    Description: '[Contains built-in filter] Convert an audio file to mp3, retaining ID3 tags, '
+module.exports.dependencies = ['import-fresh'];
+const details = () => ({
+  id: 'Tdarr_Plugin_Greg_MP3_FFMPEG_CPU',
+  Stage: 'Pre-processing',
+  Name: 'Audio Transcode to MP3 using CPU and FFMPEG',
+  Type: 'Audio',
+  Operation: 'Transcode',
+  Description: '[Contains built-in filter] Convert an audio file to mp3, retaining ID3 tags, '
       + 'and at original bitrate up to 320k - from type of: "flac,wav,ape,ogg,m4a,wma,opus" ',
-    Link: 'https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js',
-    Name: 'Audio Transcode to MP3 using CPU and FFMPEG',
-    Operation: 'Transcode',
-    Tags: 'pre-processing,ffmpeg,audio only',
-    Type: 'Audio',
-    Stage: 'Pre-processing',
-    Version: '0.0.1',
-    Inputs: [
-      {
-        name: 'codecsToInclude',
-        defaultValue: 'flac,wav,ape,ogg,m4a,wma,opus',
-        tooltip: `Codecs to exclude
+  Version: '0.0.1',
+  Tags: 'pre-processing,ffmpeg,audio only',
+  Inputs: [
+    {
+      name: 'codecsToInclude',
+      type: 'string',
+      defaultValue: 'flac,wav,ape,ogg,m4a,wma,opus',
+      inputUI: {
+        type: 'text',
+      },
+      tooltip: `Codecs to exclude
                \\nExample:\\n
                flac,wav,ape,ogg,m4a,wma,opus`,
-      },
-    ],
-  };
-}
+    },
+  ],
+});
 
 module.exports.details = details;
 
-module.exports.plugin = function plugin(file, librarySettings, inputs) {
+// eslint-disable-next-line no-unused-vars
+const plugin = (file, librarySettings, inputs, otherArguments) => {
+  // eslint-disable-next-line no-unused-vars,no-param-reassign
+  inputs = loadDefaultValues(inputs, details);
   const library = importFresh('../methods/library.js');
   const response = {
     // 320K selected over 384k intentionally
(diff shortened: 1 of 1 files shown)


Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
```diff
@@ -0,0 +1,50 @@
...
+  ${library.filters.filterByCodec(file, 'exclude', 'mp3').note}`;
+
+  if ((true && library.filters.filterByCodec(file, 'include', 'flac,wav,ape,ogg,m4a,wma,opus').outcome === true
+    && library.filters.filterByCodec(file, 'exclude', 'mp3').outcome === true)
+    || file.forceProcessing === true) {
+    response.preset = ', -map_metadata 0 -id3v2_version 3 -b:a 384k';
+    response.container = '.mp3';
+    response.handbrakeMode = false;
+    response.ffmpegMode = true;
+    response.reQueueAfter = true;
```
supersnellehenk, line 42: You can set these values as default when you first create the object.

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
```diff
@@ -0,0 +1,50 @@
...
+    infoLog: '',
+
+  };
+
+  response.infoLog += `${library.filters.filterByCodec(file, 'include', 'flac,wav,ape,ogg,m4a,wma,opus').note}
+  ${library.filters.filterByCodec(file, 'exclude', 'mp3').note}`;
+
+  if ((true && library.filters.filterByCodec(file, 'include', 'flac,wav,ape,ogg,m4a,wma,opus').outcome === true
+    && library.filters.filterByCodec(file, 'exclude', 'mp3').outcome === true)
+    || file.forceProcessing === true) {
```
supersnellehenk, line 37: You can remove the first `true &&`, doesn't do anything. You can remove all `=== true` too as that's the default.

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
```diff
@@ -0,0 +1,50 @@
...
+    response.preset = ', -map_metadata 0 -id3v2_version 3 -b:a 384k';
```
supersnellehenk, line 38: Are you sure you want 384k? That way it'll end up as a layer 2 mp3 while there's layer 3 available at 320k.

https://en.m.wikipedia.org/wiki/ISO/IEC_11172-3

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
```diff
@@ -0,0 +1,50 @@
...
+    response.container = '.mp3';
+    response.handbrakeMode = false;
+    response.ffmpegMode = true;
+    response.reQueueAfter = true;
+    response.processFile = library.actions.remuxContainer(file, 'mp3').processFile;
+    response.infoLog += library.actions.remuxContainer(file, 'mp3').note;
+    return response;
+  }
+  response.processFile = false;
```
supersnellehenk, line 47: This is the default, no need to set it :)

supersnellehenk (changes requested):

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
Greg-2600: Excellent feedback, I understand this much better now!  I changed this to 320k per the information you shared.

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
Greg-2600: Good call, removed!

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
Greg-2600: So much better now, thank you for pointing this out!

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
Greg-2600: Great call, I refactored this out and set everything during the instantiation as you recommended.

Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js
```diff
@@ -0,0 +1,44 @@
...
+
+module.exports.details = function details() {
+  return {
+    id: 'Tdarr_Plugin_Greg_MP3_FFMPEG_CPU',
+    Description: '[Contains built-in filter] Convert an audio file to mp3, retaining ID3 tags, '
+    + 'and at original bitrate up to 320k - from type of: "flac,wav,ape,ogg,m4a,wma,opus" ',
+    Link: 'https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Community/Tdarr_Plugin_Greg_MP3_FFMPEG_CPU.js',
+    Name: 'Audio Transcode to MP3 using CPU and FFMPEG',
+    Operation: 'Transcode',
+    Tags: 'pre-processing,ffmpeg,audio only,mp3',
```
supersnellehenk, line 11: I missed this, but mp3 isn't a valid tag. Valid ones are listed in the example plug-in https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Tdarr_Plugin_aaaa_Pre_Proc_Example.js

supersnellehenk (changes requested):
Other than that, looks good!

Greg-2600:
I see that import-fresh is supposed to be installed: in the node_modules. https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/node_modules/import-fresh/package.json

If there something i can do about this?

supersnellehenk:
> I see that import-fresh is supposed to be installed: in the node_modules. https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/node_modules/import-fresh/package.json
> 
> 
> 
> If there something i can do about this?

There's 3 lines you have to add, those are listed in the pre-processing example plug-in. Besides that, the checks have failed, could you fix those?

Greg-2600:
Well I added what i thought would be the 3 lines you referenced that would include the dependency the build is failing on.  It looks like the same issue exists, and I do not know what to do about it.

supersnellehenk (dismissed):

HaveAGitGat:
This is just a plugin that you’ve created through the plugin creator UI right?
 
I don’t think there’s any need to add it to the community plugins as others can create it through a few steps in the UI. Else we’d potentially end up with thousands of plugins from the plugin creator UI, all slightly different variations of each other.

Greg-2600:
> This is just a plugin that you’ve created through the plugin creator UI right?
> 
> I don’t think there’s any need to add it to the community plugins as others can create it through a few steps in the UI. Else we’d potentially end up with thousands of plugins from the plugin creator UI, all slightly different variations of each other.

Hello! Thanks for this great project.  Yes i built this in the plug-in creator UI and made some adjustment in an IDE per code review feedback.  The reason I submitted it: I noticed that many threads on Reddit and elsewhere were looking for a music collection oriented plugin that performed this feature. It’s your code base obviously and your call - I thought this was potentially scratching a common itch.

Greg-2600:
Woohoo!

:)

HaveAGitGat:
Woops lost my comment but I mentioned that I found and fixed a bug as you were using the remux action. For example an aac file in mp3 container would not get transcoded to mp3 in mp3 container.

Also can now set codecsToInclude in the library inputs:
image


Thanks :)

PR #209, opened 2021-12-13
Create parameter for new audio stream either before or after

michpas:
parameter is empty; new stream will be created after the orginal.
parameter is 'before'; new stream is created before the existing stream


HaveAGitGat:
Thanks. I fixed a windows bug and added a dropdown for the option you added:
image

michpas:
Great. thanks

PR #210, opened 2021-12-14
removed duplicated parameter setting

michpas:


HaveAGitGat:
Good catch, thanks. I've added a test for that in future.

PR #217, opened 2021-12-20
Create Tdarr_Plugin_goof1_URL_Plex_Refresh

goofballtech:
Adding PR as reference point to my Issue in the Tdarr repo. Am fine adding to community but would rather get the success/error messages added as i'd hoped.


supersnellehenk:
@goofballtech File doesn't have the `.js` extension :)

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+  Description: `Refreshes folder containing the current file in Plex so changes are picked up properly.`,
+  Version: '1.0',
+  Tags: '3rd party,post-processing,configurable',
+
+  Inputs: [
+    {
+      name: 'Url_Protocol',
+      type: 'string',
+      defaultValue: 'http',
+      inputUI: {
```
HaveAGitGat, line 19: Use dropdown menu UI here

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+
+               \\nExample:\\n
+               192.168.0.10
+
+               \\nExample:\\n
+               subdomain.domain.tld`,
+    },
+    {
+      name: 'Plex_Port',
+      type: 'string',
```
HaveAGitGat, line 49: Use number type for port

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+    response.infoLog += `☒ Above shown folder scanned in Plex! \n`
+  } else if (statusCode == 401) {
+    response.infoLog += `Plex replied that the token was not authorized on this server \n`
+  } else if (statusCode == 404) {
+    response.infoLog += `404 Plex not found, http/https is set properly? The URL used was 
+    ${redactedUrl} \n`
+  } else {
+    response.infoLog += `There was an issue reaching Plex. The URL used was 
+    ${redactedUrl} \n`
+  }
```
supersnellehenk, line 217: Strict type checking using `===` instead of `==`

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+    return response
+  }
+
+
+  let filePath = file.file
+  filePath = filePath.substring(0, filePath.lastIndexOf("/"));
+
+  if (tdarrPath || plexPath) {
+    filePath = filePath.replace(tdarrPath, plexPath)
+  }
```
supersnellehenk, line 177: Seems to me that this check should be `&&` and not `||`

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+
+  response.infoLog += `Attempting to update Plex path ${filePath} in library ${key}\n`
+
+  const fullUrl = `${type}://${url}${port ? `:${port}` : ''}/library/sections/${key}/refresh?path=${filePath}&X-Plex-Token=${token}`
+  const redactedUrl = `${type}://${url}${port ? `:${port}` : ''}/library/sections/${key}/refresh?path=${filePath}&X-Plex-Token=[redacted]`
```
supersnellehenk, line 182: You could build this as a sort of `tempUrl` variable and only append the token at the request. Now you're *basically* doing the same thing twice, just with the token or without.

Move the ternary out of the string itself to a separate variable. Increases readability.

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
supersnellehenk, line 217: Move this function out of the plugin function.

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+
+  if (!type || !url || !token || !key) {
+    response.infoLog = `Url_Protocol, Plex_Url, Plex_Token, and Library_Key are all required`
+    response.processFile = false
+    return response
+  }
+
+
+  let filePath = file.file
+  filePath = filePath.substring(0, filePath.lastIndexOf("/"));
```
supersnellehenk, line 173: Kinda redundant initial value if you're gonna override it immediately

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
```diff
@@ -0,0 +1,221 @@
...
+  const url = inputs.Plex_Url
+  const port = inputs.Plex_Port
+  const token = inputs.Plex_Token
+  const key = inputs.Library_Key
+  const plexPath = inputs.Plex_Path
+  const tdarrPath = inputs.TDarr_Path
+
+  if (!type || !url || !token || !key) {
+    response.infoLog = `Url_Protocol, Plex_Url, Plex_Token, and Library_Key are all required`
+    response.processFile = false
```
supersnellehenk, line 167: You can set `processFile` to `false` by default by having it in the response object when you first create it

supersnellehenk (changes requested):

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
goofballtech: can you tell me where the ui types are listed? I looked through the repo and didn't see what the options were so i stuck with what i found in other extensions.

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
goofballtech: it's possible that the root of one mapping is not the root of the other. the replace with then just remove the specified portion of the path.

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
goofballtech: was left over from testing, didn't realize it was still double stepping

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
goofballtech: it's already out of the plugin function.....

goofballtech: i forgot to put it back in testing to figure out why there were never getting called. Though it may be doing something crazy with the responses. Then i figured out the call backs weren't working at all and never revisited.

supersnellehenk: That's my bad. Github diff is kinda hard to read sometimes. :)

Community/Tdarr_Plugin_goof1_URL_Plex_Refresh.js
supersnellehenk: Yeah, but that's what I'm implying. The if statement checks if either isn't empty or null, thus now it could happen that someone only fills `tdarrPath` and forgets about `plexPath` (somehow) or the other way around. It'll replace `/data/something/` with an empty string, thus removing that part entirely. The other way around, it would replace nothing with something.

goofballtech: and that could also be the case. if plex is /tv and tdarr is /data/media/tv you would need to remove /data/media for the scan so plex would be plank and /data/media would be in tdarr. If tdarr is /tv and plex is /data/media/tv you would need to type / in tdarr and /data/media in plex to make the right string. i might be able to do some complex regex or make it a few else if's that series here to handle all the possible combos and some edge cases. Not sure if it'd be worth it though. Think its worth trying?

goofballtech: got bored and played for a bit. i think this fulfills your concerns with having the tdarrPath be blank but having something in plex that needs to be added. Thoughts?

```javascript
  if (tdarrPath && !plexPath) {
    filePath = filePath.replace(tdarrPath, plexPath);
  } else if (!tdarrPath && plexPath) {
    filePath = filePath.replace(/^/, plexPath);
  }
```

HaveAGitGat:
Fixed the inputUI types so should be good now

HaveAGitGat:
@goofballtech this should now be working as you expect.

PR #230, opened 2021-12-29
fix: default tag_language to empty string

### LINKED ISSUE #229: Tdarr_Plugin_MC93_Migz4CleanSubs tags tracks as english when it's disabled
`Tdarr_Plugin_MC93_Migz4CleanSubs` 

The `tag_language` option is supposed to be disabled when blank, but `loadDefaultValues` is populating that config item's default value (`eng`) when it's blank, resulting in it never being disabled. Suggest setting the default for `tag_language` to an empty string to fix.

Same issue likely applies to Tdarr_Plugin_MC93_Migz3CleanAudio but I haven't tested with a multi-audio file yet.

YodaDaCoda:
fixes this option defaulting to eng when user intends to disable

Fixes #229

### DIFF
--- a/Community/Tdarr_Plugin_MC93_Migz3CleanAudio.js
+++ b/Community/Tdarr_Plugin_MC93_Migz3CleanAudio.js
@@ -50,7 +50,7 @@ const details = () => ({
   {
     name: 'tag_language',
     type: 'string',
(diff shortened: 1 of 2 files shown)


HaveAGitGat:
Good spot, thanks.

PR #232, opened 2021-12-29
fix: remove title tag instead of setting it to ""

YodaDaCoda:
Setting `title=""` results in the title literally being `""`.

Per ffmpeg's [documentation](https://ffmpeg.org/ffmpeg-all.html):
> It is also possible to delete metadata by using an empty value

This change instead sets the title to an empty value (i.e. `title=`). Tested working for me.

Fixes #230

### DIFF
--- a/Community/Tdarr_Plugin_MC93_Migz2CleanTitle.js
+++ b/Community/Tdarr_Plugin_MC93_Migz2CleanTitle.js
@@ -111,7 +111,7 @@ const plugin = (file, librarySettings, inputs, otherArguments) => {
     )
   ) {
     try {
-      ffmpegCommandInsert += ' -metadata title="" ';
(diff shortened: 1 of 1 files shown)


HaveAGitGat (commented):
Looks good. We’ll need to check if this has any knock on effects on anything else.

YodaDaCoda:
It shouldn't cause the plugin to have to re-process any already-processed files, the existing condition [here](https://github.com/HaveAGitGat/Tdarr_Plugins/blob/2693e9abb977ab3a310b319e3774e948fd841607/Community/Tdarr_Plugin_MC93_Migz2CleanTitle.js#L106-111) should take care of it.

supersnellehenk (approved):

PR #242, opened 2022-01-04
Add cert mismatch ignore

goofballtech:
When using a custom domain the mismatched cert form the autogenerated *.plex.direct URl caused a mismatch error to print in the response instead of the actual status. Couldn't see this until the promise resolution was working as expected. Just added a parameter to ignore that since it doesn't matter for these purposes.


supersnellehenk:
Could you create a new branch instead of reusing your old one? It's causing conflicts that way. You shouldn't reuse already merged branches.

PR #244, opened 2022-01-04
ignore cert mismatch for *.plex.direct

goofballtech:
add parameter to ignore cert mismatches for custom domains since plex tries to push everything through their generated subdomains and it may see a cert issue.


HaveAGitGat:
Thanks :)

PR #262, opened 2022-01-26
Add input values and fix mkv tags

zachg99:
looks like its' merging branches and adding the updated file stats return logic. Basically two things going on, one is the variable names have been normalized and two is it does a return after updating the stat tags on the video file so that those updated stats are available to the plugin.


supersnellehenk (dismissed):

vodkapmp (dismissed):

HaveAGitGat (approved):

HaveAGitGat:
Rebased onto origin/master

PR #263, opened 2022-01-27
Bug fix for Chinese orginal language

townsste:
Bug where a Chinese original language movie would be set as und or undefined because of the return value from TMBD as 'cn'.  The iso-language expects 'zh' for Chinese.
Also the array may be unnecessary given TMDB should only every report a single language and is not reliant on imdb multiple languages.

### DIFF
--- a/Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
+++ b/Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
@@ -108,9 +108,15 @@ const processStreams = (result, file, user_langs) => {
   };
   let streamIndex = 0;
 
-  const langsTemp = result.original_language;
+  var langsTemp = result.original_language; //Change from const to var to re-define
+  
+  if (langsTemp == 'cn') langsTemp = 'zh'; //If the original language is pulled as Chinese 'cn' is used.  iso-language expects 'zh' for Chinese.
+  
   let langs = [];
 
+  /*
+  The TMDB should only every report a single language as original language and is not reliant on imdb languages.
+  
   if (Array.isArray(langsTemp)) {
     // For loop because I thought some imdb stuff returns multiple languages
     // Translates 'en' to 'eng', because imdb uses a different format compared to ffmpeg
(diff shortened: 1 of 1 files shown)


Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
```diff
@@ -120,7 +126,15 @@ const processStreams = (result, file, user_langs) => {
   } else {
     langs.push(languages.alpha2ToAlpha3B(langsTemp));
   }
-
+  */
+  langs.push(languages.alpha2ToAlpha3B(langsTemp));
+  
+  //Some console reporting for clarification of what the plugin is using and reporting.
```
supersnellehenk, line 132: Unnecessary comment

supersnellehenk (changes requested):

Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
```diff
@@ -108,9 +108,15 @@ const processStreams = (result, file, user_langs) => {
...
   let streamIndex = 0;
 
-  const langsTemp = result.original_language;
+  var langsTemp = result.original_language; //Change from const to var to re-define
+  
+  if (langsTemp == 'cn') langsTemp = 'zh'; //If the original language is pulled as Chinese 'cn' is used.  iso-language expects 'zh' for Chinese.
+  
   let langs = [];
 
+  /*
```
supersnellehenk, line 117: Instead of commenting it out you can just remove it. Apparently I got `original_language` mixed up with `spoken_languages` somewhere in development. https://developers.themoviedb.org/3/movies/get-movie-details

Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
```diff
@@ -120,7 +126,15 @@ const processStreams = (result, file, user_langs) => {
...
   }
-
+  */
+  langs.push(languages.alpha2ToAlpha3B(langsTemp));
+  
+  //Some console reporting for clarification of what the plugin is using and reporting.
+  response.infoLog += 'Original language: ';
+  response.infoLog += `${langsTemp}`;
+  response.infoLog += ', Using code: ';
+  response.infoLog += `${languages.alpha2ToAlpha3B(langsTemp)}\n`;
```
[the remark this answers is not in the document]
townsste: I attempted this last night and tdarr decided it did not like a single += and kept giving me a read error, but looks like tdarr is fine with it now.

Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
townsste: Did not know if you wanted to keep the loop so thus the comment out.

supersnellehenk (approved):

PR #265, opened 2022-02-01
Add QSV HEVC plugin

Boosh1:
This plugin is for encoding files to HEVC/H265 using QSV via ffmpeg. It uses QSV specific ffmpeg commands rather than vaapi. (i.e cmd contains: "-hwaccel qsv -c:v hevc_qsv"). I don't _think_ any of the other current QSV plugins do that.

Has multiple options to allow for tighter control of bitrate on encode (encoder pre-set speed, min average bitrate & max average bitrate) & also contains logic to allow the reprocessing of files already in hevc, av1 or vp9. This is disabled by default but when enabled will allow files to be crunched down if file size & quality needed to be better controlled. Extra bitrate controls like the cut-off should prevent the file from looping forever.

Was made for my purposes but want to share with the community.

### DIFF
--- a/Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
+++ b/Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
@@ -1,0 +1,399 @@
+// All credit for original plugin logic goes to Migz.
+// This Plugin is essentially just his NVENC/CPU plugin modified to work with QSV & with extra hevc logic.
+// Extra logic is mainly to control encoder quality/speed & to allow HEVC files to be reprocessed to reduce file size
+// NOTE - This does not use VAAPI, it is QSV only. So newer intel igpus only. 8th+ gen should work.
+function details() {
+  return {
+    id: 'Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC',
+    Stage: 'Pre-processing',
+    Name: 'Boosh-Transcode using QSV GPU & FFMPEG',
+    Type: 'Video',
+    Operation: 'Transcode',
+    Description: `This is a QSV specific plugin. An INTEL QSV enabled CPU is required. 8th+ gen should work.
+    Files not in H265/HEVC will be transcoded into H265/HEVC using Quick Sync Video (QSV) via Intel GPU with ffmpeg.
+    Settings are dependant on file bitrate working by the logic that H265 can support the same ammount of data at 
+    half the bitrate of H264. This plugin will skip files already in hevc, av1 & vp9 unless "reconvert_hevc" is 
+    marked as true. If it is then these will be reconverted again into hevc if they exceed the bitrate specified 
+    in "hevc_max_bitrate". Reminder! An INTEL QSV enabled CPU is required.`,
+    Version: '1.0',
+    Tags: 'pre-processing,ffmpeg,video only,qsv,h265,hevc,configurable',
+    Inputs: [{
+      name: 'container',
+      type: 'string',
+      defaultValue: 'slower',
+      inputUI: {
+        type: 'dropdown',
+        options: [
+          'mkv',
+          'mp4',
+        ],
+      },
+      tooltip: `Specifies the output container of the file
+      \\n Ensure that all stream types you may have are supported by your chosen container.
+      \\n MKV is recommended.
+      \\nExample:\\n
+      mkv
+      \\nExample:\\n
+      mp4`,
+    },
+    {
+      name: 'encoder_speedpreset',
+      type: 'string',
+      defaultValue: 'slower',
+      inputUI: {
+        type: 'dropdown',
+        options: [
+          'veryfast',
+          'faster',
+          'fast',
+          'medium',
+          'slow',
+          'slower',
+          'veryslow',
+        ],
+      },
+      tooltip: `Specify the encoder speed/preset to use. 
+      Slower options mean slower encode but better quality and faster have quicker encodes but worse quality.
+      \\n Default is "slower". For almost best quality with faster encode`,
+    },
+    {
+      name: 'enable_10bit',
+      type: 'boolean',
+      defaultValue: false,
+      inputUI: {
+        type: 'dropdown',
+        options: [
+          'false',
+          'true',
+        ],
+      },
+      tooltip: `Specify if we want to enable 10bit encoding. 
+      If this is enabled 10bit files will be processed and converted into 10bit 
+      hevc using main10 profile and with p010le pixel format.`,
+    },
+    {
+      name: 'bitrate_cutoff',
+      tooltip: `Specify bitrate cutoff, files with a current bitrate lower then this will not be processed.
+      \\n Rate is in kbps.
+      \\n Leave empty to disable cutoff.
+      \\nExample:\\n
+      2500
+      \\nExample:\\n
+      1500`,
+    },
+    {
+      name: 'max_average_bitrate',
+      tooltip: `Specify a maximum average bitrate, when encoding we take the current bitrate and halve it 
+      to get an average target. This option sets a upper limit to that average 
+      (i.e if you have a video bitrate of 10000, half is 5000, if your maximum desired average bitrate is 4000
+      then we use that as the target instead of 5000).
+      \\n Rate is in kbps.
+      \\n Leave empty to ignore.
+      \\nExample:\\n
+      2000
+      \\nExample:\\n
+      1000`,
+    },
+    {
+      name: 'min_average_bitrate',
+      tooltip: `Specify a minimum average bitrate, when encoding we take the current bitrate and halve it. 
+      This option sets a lower limit to that average (i.e if you have a video bitrate of 3000, 
+      half is 1500, if your minimum desired average bitrate is 2000 then we use that as the target instead of 1500).
+      \\n Rate is in kbps.
+      \\n Leave empty to ignore.
+      \\nExample:\\n
+      2000
+      \\nExample:\\n
+      1000`,
+    },
+    {
+      name: 'reconvert_hevc',
+      type: 'boolean',
+      defaultValue: false,
+      inputUI: {
+        type: 'dropdown',
+        options: [
+          'false',
+          'true',
+        ],
+      },
+      tooltip: `Specify if we want to reprocess HEVC, vp9 or av1 files 
+      (i.e reduce bitrate of files already in those codecs).NOT reccomeneded to use so leave blank if unsure. 
+      NEEDS to be used in conjuntion with "bitrate_cutoff" otherwise is ignored.
+      \\n Will allow files that are already HEVC, vp9 or av1 to be reprocessed
+      \\n Useful in certain situations, perhaps you have a file which is HEVC 
+      but extremely high bitrate and you'd like to reduce
+      \\n\\n WARNING!! IF YOU HAVE VP9 OR AV1 FILES YOU WANT TO KEEP IN THOSE FORMATS THEN DO NOT USE THIS OPTION
+      \\n
+      \\nExample:\\n
+      true
+      \\nExample:\\n
+      false`,
+    },
+    {
+      name: 'hevc_max_bitrate',
+      tooltip: `Specify a maximum allowed average bitrate for hevc or similar files. 
+      This option is to be used if you want to ensure hevc files don't exceed a set bitrate.
+      \\n Rate is in kbps.
+      \\n If empty we will take the bitrate_cutoff and multiply x2 for a safe limit
+      \\nExample:\\n
+      4000
+      \\nExample:\\n
+      3000`,
+    },
+    ],
+  };
+}
+
+function plugin(file, librarySettings, inputs) {
+  const response = {
+    processFile: false,
+    preset: '',
+    handBrakeMode: false,
+    FFmpegMode: true,
+    reQueueAfter: true,
+    infoLog: '',
+  };
+
+  let duration = '';
+
+  // Set up required variables.
+  let videoIdx = 0;
+  let extraArguments = '';
+  let bitrateSettings = '';
+  let inflatedCutoff = '';
+  let currentBitrate = '';
+  let targetBitrate = '';
+  let minimumBitrate = '';
+  let maximumBitrate = '';
+  let main10 = '';
+
+  // Check if inputs.container has been configured. If it hasn't then exit plugin.
+  if (inputs.container === '') {
+    response.infoLog += '☒ Plugin has not been configured, please configure required options. Skipping this plugin. \n';
+    response.processFile = false;
+    return response;
+  }
+  response.container = `.${inputs.container}`;
+
+  // Check if file is a video. If it isn't then exit plugin.
+  if (file.fileMedium !== 'video') {
+    response.processFile = false;
+    response.infoLog += 'File is not a video. \n';
+    return response;
+  }
+
+  // Check if duration info is filled, if so times it by 0.0166667 to get time in minutes.
+  // If not filled then get duration of stream 0 and do the same.
+  if (typeof file.meta.Duration !== 'undefined') {
+    duration = file.meta.Duration * 0.0166667;
+  } else {
+    duration = file.ffProbeData.streams[0].duration * 0.0166667;
+  }
+
+  // Work out currentBitrate using "Bitrate = file size / (number of minutes * .0075)"
+  // Used from here https://blog.frame.io/2017/03/06/calculate-video-bitrates/
+  currentBitrate = (file.file_size / (duration * 0.0075));
+  // Use the same calculation used for currentBitrate but divide it in half to get targetBitrate.
+  // Logic of h265 can be half the bitrate as h264 without losing quality.
+  targetBitrate = (file.file_size / (duration * 0.0075) / 2);
+  // Allow some leeway under and over the targetBitrate.
+  minimumBitrate = (targetBitrate * 0.75);
+  maximumBitrate = (targetBitrate * 1.25);
+
+  response.infoLog += `☑ It looks like the current bitrate is ${currentBitrate}. \n`;
+
+  // If targetBitrate or currentBitrate comes out as 0 then something
+  // has gone wrong and bitrates could not be calculated.
+  // Cancel plugin completely.
+  if (targetBitrate <= 0 || currentBitrate <= 0) {
+    response.processFile = false;
+    response.infoLog += '☒ Target bitrate could not be calculated. Skipping this plugin. \n';
+    return response;
+  }
+
+  if (inputs.reconvert_hevc === 'true' && inputs.bitrate_cutoff === '') {
+    response.processFile = false;
+    response.infoLog += `☒ Reconvert HEVC is ${inputs.reconvert_hevc}, however there is no bitrate cutoff set 
+    so we have no way to know when to stop processing this file. 
+    Either set reconvert_HEVC to false or set a bitrate cutoff. Skipping this plugin. \n`;
+    return response;
+  }
+
+  // Check if inputs.bitrate cutoff has something entered.
+  // (Entered means user actually wants something to happen, empty would disable this).
+  if (inputs.bitrate_cutoff !== '') {
+    // Checks if currentBitrate is below inputs.bitrate_cutoff.
+    // If so then cancel plugin without touching original files.
+    if (currentBitrate <= inputs.bitrate_cutoff) {
+      response.processFile = false;
+      response.infoLog += `☑ Current bitrate is below set cutoff of ${inputs.bitrate_cutoff}. Cancelling plugin. \n`;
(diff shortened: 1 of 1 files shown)


Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+          }
+          // If we're not using the hevc max bitrate then we need a safety net
+          // to try and ensure we don't keep looping this plugin.
+        } else if (currentBitrate > (inputs.bitrate_cutoff * 2)) {
+          inflatedCutoff = Math.round((inputs.bitrate_cutoff * 2));
+          response.infoLog += `☒ Reconvert_hevc is ${inputs.reconvert_hevc} & the file is already hevc, vp9 or av1. 
+          hevc specific cutoff not set so bitrate_cutoff is multiplied by 2 for safety! 
+          Cutoff now temporarily ${inflatedCutoff}. \n The file is still above this new cutoff! Reconverting. \n\n`;
+        } else {
+          response.processFile = false;
```
supersnellehenk, line 328: Unnecessary processFile

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+    // Are we encoding 10 bit files?
+    if (inputs.enable_10bit === 'true') {
+      // Check if video stream is 10bit. Currently set profile main10 and pixel format.
+      // Still need to test if this really works correctly
+      if (file.ffProbeData.streams[i].profile === 'High 10'
+        || file.ffProbeData.streams[i].bits_per_raw_sample === '10') {
+        main10 = '-profile:v main10 -pix_fmt p010le';
+        response.infoLog += '☑ File is 10bit or HDR. Setting Main10 Profile & pixel format \n\n';
+      }
+    }
```
supersnellehenk, line 347: What if people want to convert from 8 bit to 10 bit? If you don't set these afaik it'll just copy the data from the source. HDR will get lost this way as the pixel format is different.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+
+    // Increment videoIdx.
+    videoIdx += 1;
+  }
+
+  // Set bitrateSettings variable using bitrate information calulcated earlier.
```
supersnellehenk, line 353: calculated*

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+      response.preset = '-hwaccel qsv -c:v mpeg1';
+    } else if (file.video_codec_name === 'mpeg2') {
+      response.preset = '-hwaccel qsv -c:v mpeg2_qsv';
+    } else if (file.video_codec_name === 'vc1') {
+      response.preset = '-hwaccel qsv -c:v vc1_qsv';
+    } else if (file.video_codec_name === 'vp8') {
+      response.preset = '-hwaccel qsv -c:v vp8_qsv';
+    } else if (file.video_codec_name === 'av1') {
+      response.preset = '-hwaccel qsv -c:v av1';
+    }
```
supersnellehenk, line 387: Switch > this if else tree

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+          \n The file is NOT above this new cutoff. Exiting plugin. \n\n`;
+          return response;
+        }
+      }
+    }
+
+    // Are we encoding 10 bit files?
+    if (inputs.enable_10bit === 'true') {
+      // Check if video stream is 10bit. Currently set profile main10 and pixel format.
+      // Still need to test if this really works correctly
```
supersnellehenk, line 341: If it needs testing, why is it in a production plugin? :)

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+  // Set bitrateSettings variable using bitrate information calulcated earlier.
+  bitrateSettings = `-b:v ${targetBitrate}k -minrate ${minimumBitrate}k `
+    + `-maxrate ${maximumBitrate}k -bufsize ${currentBitrate}k`;
+  // Print to infoLog information around file & bitrate settings.
+  response.infoLog += `\nContainer for output selected as ${inputs.container}. \n`;
+  response.infoLog += `The current file bitrate = ${currentBitrate} \n`;
+  response.infoLog += 'Bitrate settings: \n';
+  response.infoLog += `Target = ${targetBitrate} \n`;
+  response.infoLog += `Minimum = ${minimumBitrate} \n`;
+  response.infoLog += `Maximum = ${maximumBitrate} \n`;
```
supersnellehenk, line 362: In the bitrate settings they're suffixed with a `k`, but here they're not.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+    } else if (file.video_codec_name === 'vc1') {
+      response.preset = '-hwaccel qsv -c:v vc1_qsv';
+    } else if (file.video_codec_name === 'vp8') {
+      response.preset = '-hwaccel qsv -c:v vp8_qsv';
+    } else if (file.video_codec_name === 'av1') {
+      response.preset = '-hwaccel qsv -c:v av1';
+    }
+  }
+
+  response.preset += `,-map 0 -c:v hevc_qsv ${bitrateSettings} `
```
supersnellehenk, line 390: instead of `,` use `<io>`

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+    + `-preset ${inputs.encoder_speedpreset} -look_ahead 1 ${main10} 
+    -c:a copy -c:s copy -max_muxing_queue_size 9999 ${extraArguments}`;
```
supersnellehenk, line 392: Plugin will fail if there's no audio or subtitles.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+          response.infoLog += `☑ File is already hevc or vp9 & in ${inputs.container}. \n`;
+          return response;
+        }
+        // Check if codec of stream is hevc, vp9 or av1
+        // AND check if file.container does NOT match inputs.container.
+        // If so remux file.
+        if ((file.ffProbeData.streams[i].codec_name === 'hevc' || file.ffProbeData.streams[i].codec_name === 'vp9'
+          || file.ffProbeData.streams[i].codec_name === 'av1') && file.container !== inputs.container) {
+          response.infoLog += `☒ File is hevc or vp9 but is not in ${inputs.container} container. Remuxing. \n`;
+          response.preset = `, -map 0 -c copy ${extraArguments}`;
```
supersnellehenk, line 296: `<io>` over `,`

supersnellehenk: This would probably need extra logic for streams that aren't compatible between mp4 and mkv, like `mov_text`

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+    }
+  }
+
+  // Check if inputs.min_average_bitrate has something entered.
+  // (Entered means user actually wants something to happen, empty would disable this).
+  if (inputs.min_average_bitrate !== '') {
+    // Checks if inputs.bitrate_cutof is below inputs.min_average_bitrate.
+    // If so then set currentBitrate to the minimum allowed.)
+    if (inputs.bitrate_cutoff < inputs.min_average_bitrate) {
+      response.processFile = false;
```
supersnellehenk, line 257: Unnecessary processFile

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+      \\n MKV is recommended.
+      \\nExample:\\n
+      mkv
+      \\nExample:\\n
+      mp4`,
+    },
+    {
+      name: 'encoder_speedpreset',
+      type: 'string',
+      defaultValue: 'slower',
```
supersnellehenk, line 42: Not sure about this, but does `slower` provide any significant savings/quality compared to `slow`?

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+      defaultValue: false,
+      inputUI: {
+        type: 'dropdown',
+        options: [
+          'false',
+          'true',
+        ],
+      },
+      tooltip: `Specify if we want to reprocess HEVC, vp9 or av1 files 
+      (i.e reduce bitrate of files already in those codecs).NOT reccomeneded to use so leave blank if unsure.
```
supersnellehenk, line 121: recommended

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+// Extra logic is mainly to control encoder quality/speed & to allow HEVC files to be reprocessed to reduce file size
+// NOTE - This does not use VAAPI, it is QSV only. So newer intel igpus only. 8th+ gen should work.
+function details() {
+  return {
+    id: 'Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC',
+    Stage: 'Pre-processing',
+    Name: 'Boosh-Transcode using QSV GPU & FFMPEG',
+    Type: 'Video',
+    Operation: 'Transcode',
+    Description: `This is a QSV specific plugin, VAPPI is NOT used. So an INTEL QSV enabled CPU is required.
```
supersnellehenk, line 12: VAAPI

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+  return {
+    id: 'Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC',
+    Stage: 'Pre-processing',
+    Name: 'Boosh-Transcode using QSV GPU & FFMPEG',
+    Type: 'Video',
+    Operation: 'Transcode',
+    Description: `This is a QSV specific plugin, VAPPI is NOT used. So an INTEL QSV enabled CPU is required. 
+    8th+ gen should work. Files not in H265/HEVC will be transcoded into H265/HEVC using Quick Sync Video (QSV) 
+    via Intel GPU with ffmpeg. Settings are dependant on file bitrate working by the logic that H265 can support 
+    the same ammount of data at half the bitrate of H264. This plugin will skip files already in HEVC, AV1 & VP9
```
supersnellehenk, line 15: amount

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
+// All credit for original plugin logic goes to Migz.
+// This Plugin is essentially just his NVENC/CPU plugin modified to work with QSV & with extra hevc logic.
+// Extra logic is mainly to control encoder quality/speed & to allow HEVC files to be reprocessed to reduce file size
+// NOTE - This does not use VAAPI, it is QSV only. So newer intel igpus only. 8th+ gen should work.
```
supersnellehenk, line 4: Is the driver support up to snuff now for the containers? Plus, how would people use this on Linux bare metal?

supersnellehenk (changes requested):

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: Driver support is now fixed in containers. - https://github.com/HaveAGitGat/Tdarr/issues/452
FFMPEG in the container is using jellyfins version now which has QSV support packaged in ready to use. So container users just need a valid plugin and the right hardware to leverage QSV.
As for bare metal users, assuming Tdarr packages jellyfin ffmpeg there as well then they should just need to install up to date drivers and that should work. I can't verify that though.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+      \\nExample:\\n
+      4000
+      \\nExample:\\n
+      3000`,
+    },
+    ],
+  };
+}
+
+function plugin(file, librarySettings, inputs) {
```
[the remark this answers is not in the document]
Boosh1: I believe this is done correctly now

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+      // Still need to test if this really works correctly
+      if (file.ffProbeData.streams[i].profile === 'High 10'
+        || file.ffProbeData.streams[i].bits_per_raw_sample === '10') {
+        main10 = '-profile:v main10 -pix_fmt p010le';
+        response.infoLog += '☑ File is 10bit or HDR. Setting Main10 Profile & pixel format \n\n';
+      }
+    }
+
+    // Increment videoIdx.
+    videoIdx += 1;
```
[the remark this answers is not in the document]
Boosh1: Whoops. This was in the wrong place so moved back inside the if check of the loop. We need it to find the ID for the video streams specifically for the first function there which will remove png and mjpeg streams

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
```diff
@@ -0,0 +1,400 @@
...
+        } else {
+          response.processFile = false;
+          inflatedCutoff = Math.round((inputs.bitrate_cutoff * 2));
+          response.infoLog += `☑ Reconvert_hevc is ${inputs.reconvert_hevc} & the file is already hevc, vp9 or av1 
+          so bitrate_cutoff is multiplied by 2! Cutoff now temporarily ${inflatedCutoff}. 
+          \n The file is NOT above this new cutoff. Exiting plugin. \n\n`;
+          return response;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
Boosh1: Should be clearer now.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: Replaced with a switch

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: Good catch. Overhauled this now. 
If this is set to true then we convert to 10 bit. Then the rest of the check is just for catching if the file is already 10 bit where we just disable hardware decode to avoid issues.
As for HDR. That now has a separate check which if found we then exit the plugin. HDR seems to be a bit involved and outside the scope of this plugin tbh.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: Was mainly referring to HDR being untested rather than 10 bit. Now there is a HDR check which is tested and should catch HDR files.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: Added in migz `force_conform` logic which should allow the encode to finish by dropping thigns like mov_text etc.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: This is not what I've experienced. I've put through video only files and the encode completed fine with no error. I believe this cmd argument will copy audio and subtitles if they are found, but won't error if none are present.

Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
Boosh1: Hard to say. Intel did a white paper regarding QSV on ffmpeg performance. - https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/cloud-computing-quicksync-video-ffmpeg-white-paper.pdf
They describe `slower` as close to best quality with faster performance. `Slow` is just described as between `slower` and `medium`.
Regardless I've defaulted this to `slow` now since I suspect that's the profile most would prefer, and the user can change this if they want to. Also added this link into the script so user can view it and make their own decision.

Boosh1:
I think all issues should be resolved now.

HaveAGitGat:
A bunch of inputs are missing type, defaultValue and inputUI so need to resolve those.

`Plugin Input keys are not in correct order: './Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js' : bitrate_cutoff`

dalcombright:
seeing the following when I grabbed the latest:

`[NULL @ 0x55a7ea7c68c0] Unable to find a suitable output format for '/temp/MOVIE (YYYY).bluray.remux.1080p-TdarrCacheFile-ZPRuqqW6J.slower'
/temp/MOVIE (YYYY).bluray.remux.1080p-TdarrCacheFile-ZPRuqqW6J.slower: Invalid argument
Tdarr ALERT: NO OUTPUT FILE PRODUCED:
/temp/MOVIE (YYYY).bluray.remux.1080p-TdarrCacheFile-ZPRuqqW6J.slower`

Boosh1:
My bad, didn't realise those were missing. That should be fixed now.

@dalcombright - Thanks for testing this. I think you're on the wrong version though, that error was present in my very first commit on my branch, but I fixed it (accidently defaulted the container to "slower"). Grab the latest and it should be fine.

HaveAGitGat:
Looks good, only issue is some of the inputs such as 'bitrate_cutoff' should probably be `number` type as they are numbers instead of strings.

Boosh1:
Changed the bitrate settings input type from string to number as suggested.

Boosh1:
Realised that my change to numbers broke a few things since blank inputs i.e `''` would end up meaning 0. Done a fix for that now as well.

Boosh1:
Changed the numbers back to strings. While being set to numbers worked, it meant we'd need to default them to `0` and not a blank entry `''`. Users seeing `0` already entered in on their plugin would just cause confusion I think. 
Other plugins seem to have bitrate entries as strings as well so seems consistent and it's fully functional with them being strings.

HaveAGitGat:
The other plugins have them as strings because they were created before the inputs had types, so as a quick upgrade had the type set to string, but they should be numbers. 

Simply changing `Leave empty to disable cutoff` to `Leave  as 0 to disable cutoff` would suffice but open to suggestions.

Boosh1:
If they should be number then that's fair enough, and the tooltip update seems sensible. I've changed this again and updated the tooltip to explain.

HaveAGitGat:
Looks goods, thanks.

HaveAGitGat (approved):

supersnellehenk (approved):

PR #267, opened 2022-02-05
Better precision for checking file duration

Enigma0:
Lets you set upper and lower bounds more accurately and prints out more accurate error figures. Default has error range of 2%, this default has error range of 1% (.5% above or below).

In testing it appears that going even to 99.8 <> 100.2 would also be acceptable and present little to no false positives with an error range of .4%. I've seen random failures that are just barely caught at 98% of original.

newData & oldData maintain up to 3 to 5 decimal places of accuracy sometimes. This allows ratio to more accurately reflect that. 3 places should be sufficient.


HaveAGitGat:
Looks good

HaveAGitGat (approved):

HaveAGitGat:
Workflow tests didn't run for some reason so bumped the version to re-trigger it.

PR #268, opened 2022-02-11
Prevent needless tmdb api calls

townsste:
This modification adds a kill plugin command based off of henk_Add_Specific_Audio_Codec plugin.

With Tdarr's cyclic processing this plugin will always check and potentially run an API call to tmdb.  

This is used to process the plugin once.
Then check if the plugin has been processed by checking for copyright metadata set within the audio.
If true plugin will terminate.
If false plugin will continue.


townsste:
Potential infinite transcode loop found.

PR #271, opened 2022-02-17
Fix error with ts files and cleanup

zachg99:
Fixed an out of memory issue that prevented .ts files from being
processed by removing the framerate portion of the video filter
command when processing a .ts file.

Removed a bunch of variables used for building the ffmpeg command and
moved the actual strings down to where the logic is used. Reducing
plugin file size and variable ussage.

Fixed a minor bug related to changing the video height where it would
always use the 1080 value rather than using the value used in the input.


HaveAGitGat:
Gave it a run and getting this error:

`2022-03-19T16:25:40.645Z os8QNsFH4xa:Node[Desktop]:Worker[oNTofzVzo]:Error TypeError: Cannot read property 'FromStats_BitRate' of undefined`

zachg99:
Doh!

So, I get that error only if the mediainfo tags are missing. That’s why I run the mkvpropedit --add-track-statistics-tags "${currentFileName}” the first time through to set those tags as many files don’t have them or they have incorrect information.

Now, I’m only calling that scan if the stats on the file are out of date which is set by the stats_days input parameter.

Is it possible this file is missing the bitrate tags and yet has a recent stats date on the file you’re testing with?

--
Our problems are like diamonds. We inflate their value because we don't understand how many they have in Africa.

supersnellehenk (changes requested):
General note: remove all the unnecessary comments. Just makes it harder to read and you can always revert using git.

zachg99:
Will do.

--
Our problems are like diamonds. We inflate their value because we don't understand how many they have in Africa.

supersnellehenk (approved):

vodkapmp (approved):

buroa:
> 

Hey @zachg99 -- are you running into the problem of constantly re-queuing files in h.264 because the plugin skips the transcode due to this line: https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Community/Tdarr_Plugin_JB69_JBHEVCQSV_MinimalFile.js#L338

I can't be the only one here. One the second run, it transcodes to h.265.

zachg99:
I haven’t seen that happen but I haven’t been transcoding a ton lately (mainly .ts files fro plex dvr recordings).

If you have the latest version of Tdarr that tool is built into it and is used to update/correct the mediainfo tags on the file. I was noticing a lot that were missing bitrate info so this should be scanning the file and all the mediainfo files as appropriate. It takes a while to run, especially with larger files, but should correct any issues then get re-processed by the plugin. Upon the second round of processing it uses the updated file object that then has all the mediainfo tags. That’s the idea of the return right after it. 

I wonder if it needs to return something else to tell Tdarr to re-process the same plugin rather than moving on in the stack.

buroa:
@zachg99 What tool exactly is inside Tdarr to update these? I can requeue health checks/etc but it doesn't appear to update the tags.

zachg99:
It's that command it's running: mkvpropedit --add-track-statistics-tags 

HGG added the tool with the 2.00.14 release. Prior to that point i was manually installing it from within the container.

mkvproedit is part of mkvtoolnix tools package. The plugin does a check on the stats date and if they were updated within the last so many days (default is 21 but it can be changed as an input value) it runs the above command.

buroa:
@zachg99 Ah, right. I am using the latest .16 version of Tdarr. The plugin definitely runs to modify the tags and then skips the rest the "transcoding" part of the plugin.

zachg99:
So, that's by design, but what isn't is the plugin doesn't get called a second time. The reason for exiting to rerun is to refresh the file object information with the new tags. I was under the impression that it would get called a 2nd time by the diagram here:  https://docs.tdarr.io/docs/plugins/basics/

I wonder if i should pull that logic out into it's own plugin, set to run as the first item in the plugin stack.

buroa:
@zachg99 I tried to build it into it's own plugin as well but it still wasn't working. It seems like Tdarr plugin stack doesn't actually update the metadata throughout it's runtime. So example, if one plugin updated the metadata: the new metadata would not be available to the next plugin because Tdarr is basically caching that stuff throughput the flow.

cc @HaveAGitGat

zachg99:
If there were a way to tell tdarr to re-scan the file tags within the plugin, I could call that instead of the return and that should resolve this issue.

buroa:
> If there were a way to tell tdarr to re-scan the file tags within the plugin, I could call that instead of the return and that should resolve this issue.

Yeah that would work. I'm not sure how to tell it to update the tags, although.

zachg99:
The alternative is to comment out the return that's just after that tag update.  Many files have bitrate in the tags already, just some don't. If it doesn't have a tag it will error and you'll need to requeue it. It should work the second time as you've already updated the tags the first time through.

PR #276, opened 2022-03-10
Fix sonarr dir match

supersnellehenk:
Not everyone has the same paths between Sonarr and Tdarr, especially when running in Docker. This fixes the issue where those paths don't match by just looking at the show folder.


Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-await-in-loop */
-module.exports.dependencies = ['axios', '@cospired/i18n-iso-languages', 'path'];
```
HaveAGitGat, line 2: path is part of the nodejs standard library so no need to install it

HaveAGitGat (changes requested):

Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
supersnellehenk: I removed it, you're replying on the old version :)

HaveAGitGat (approved):

PR #278, opened 2022-03-12
Also Remove any Subs tagged with SDH

Fish2:
Add removal of SDH Subtitle to work the same and in line with "[Tdarr_Plugin_MC93_Migz3CleanAudio.js](https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Community/Tdarr_Plugin_MC93_Migz3CleanAudio.js)"


HaveAGitGat:
Thanks

PR #280, opened 2022-03-18
Add 10bit output boolean

argonan0:
Certain media benefits from being pushed into 10bit if not already. This could be improved by detecting if the source file already is 10bit and maybe assigning a special main10 profile.

### DIFF
--- a/Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
+++ b/Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
@@ -126,6 +126,24 @@ const details = () => ({
         \\nExample:\\n
         true`,
     },
+    {
+    name: 'enable10bit',
+    type: 'boolean',
+    defaultValue: false,
+    inputUI: {
+      type: 'dropdown',
+      options: [
+        'false',
+        'true',
+      ],
+    },
+    tooltip: `Specify if output file should be 10bit. Default is false.
+                    \\nExample:\\n
+                    true
+
+                    \\nExample:\\n
+                    false`,
+    },
   ],
 });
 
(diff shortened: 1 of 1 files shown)


Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -222,8 +240,16 @@ const plugin = (file, librarySettings, inputs, otherArguments) => {
 
   // encoding settings
   response.preset += `,-map 0 -dn -c:v libx265 -preset ${ffmpegPreset}`
-  + ` -x265-params crf=${crf}:bframes=${bframe}:rc-lookahead=32:ref=6:b-intra=1:aq-mode=3`
-  + ' -a53cc 0 -c:a copy -c:s copy -max_muxing_queue_size 9999';
+  response.preset += ` -x265-params crf=${crf}:bframes=${bframe}:rc-lookahead=32:ref=6:b-intra=1:aq-mode=3`
+  response.preset += ' -a53cc 0 -c:a copy -c:s copy';
```
supersnellehenk, line 244: Why did you split this instead of just concating to the original string?

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -222,8 +240,16 @@ const plugin = (file, librarySettings, inputs, otherArguments) => {
...
+
+  // Check if 10bit variable is true.
+  if (inputs.enable10bit === true) {
+    // If set to true then add 10bit argument
+    response.preset += ' -pix_fmt p010le';
+  }
```
supersnellehenk, line 250: Can be simplified to `response.preset += inputs.enable10bit ? ' -pix_fmt p010le' : ''`.

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
```diff
@@ -126,6 +126,24 @@ const details = () => ({
...
+        'false',
+        'true',
+      ],
+    },
+    tooltip: `Specify if output file should be 10bit. Default is false.
+                    \\nExample:\\n
+                    true
+
+                    \\nExample:\\n
+                    false`,
```
supersnellehenk, line 145: Should probably add this forces 10 bit. If disabled it takes whatever the bit depth of the original file is.

supersnellehenk (changes requested):

Community/Tdarr_Plugin_vdka_Tiered_CPU_CRF_Based_Configurable.js
argonan0: I'm not sure of the proper syntax here for multiline. This seemed safer/ more clear.

supersnellehenk: > I'm not sure of the proper syntax here for multiline. This seemed safer/ more clear.

The way it was before was fine.

argonan0 (commented):
should be fixed

argonan0 (commented):
Fixed the rename issue

supersnellehenk (approved):

vodkapmp (approved):

PR #286, opened 2022-03-24
PCM (WAV) to FLAC plugin

aunefyren:
Very much like ALAC codecs being lossless-ly converted to FLAC, as can WAV if the bit-depth is below 32. This plugin looks for PCM codecs (used in WAV) and transcodes to FLAC if the bit-depth is 16 or 24.

### DIFF
--- a/Community/Tdarr_Plugin_fu85_aune_pcm_to_flac.js
+++ b/Community/Tdarr_Plugin_fu85_aune_pcm_to_flac.js
@@ -1,0 +1,54 @@
+/* eslint-disable */
+const details = () => {
+  return {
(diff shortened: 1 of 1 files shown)


supersnellehenk:
So, instead of multiple plugins that _basically_ achieve the same output for different input files, could we perhaps merge them together? Give the user to option to convert both wav and alac, or only one of the two?

aunefyren:
> So, instead of multiple plugins that _basically_ achieve the same output for different input files, could we perhaps merge them together? Give the user to option to convert both wav and alac, or only one of the two?

Great idea! Made an update to my original plugin

PR #296, opened 2022-04-23
Added Tdarr_Plugin_MARN_Migz5ConvertAudio plugin

Marnalas:
Does the same thing as Tdarr_Plugin_MC93_Migz5ConvertAudio except that it looks at languages.

Examples :
Tdarr_Plugin_MC93_Migz5ConvertAudio
  audio track 6 channels in english + audio track 2 channels in french => does nothing
Tdarr_Plugin_MARN_Migz5ConvertAudio
  audio track 6 channels in english + audio track 2 channels in french => creates an audio track 2 channels in english


supersnellehenk:
Could we instead of creating a new plugin implement this in the Migz plugin perhaps? Otherwise, we get a lot of plugins that basically do the same thing/have the same code but with a slight change. Then create a toggle to be 'language aware' and default that to false to preserve the original logic of the plugin.

Marnalas:
No problem, I think your proposition is better. I'm gonna do that.
Thank you for your answer. I'll close this request and create a new one once I'm done.

PR #299, opened 2022-04-25
Prettier setup

supersnellehenk:
https://prettier.io/docs/en/why-prettier.html

ESLint isn't strict enough regarding style in my opinion. This solves the debate about what to do in some cases.

Adding a pre-commit hook to have contributors lint their files before hitting the CI, saving having to go back and forth. Especially for new contributors which need approval before the CI runs.

I haven't touched the eslint disabled plugins, as those require a lot of manual work to convert. FYI, they threw about 1000 errors all together spread over 51 files.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,31 +1,43 @@
 {
-    "env": {
-        "commonjs": true,
-        "es6": true,
-        "node": true
-    },
-    "extends": [
-        "airbnb-base"
+  "env": {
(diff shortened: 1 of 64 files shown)


HaveAGitGat:
@supersnellehenk will close this as a bit outdated. Was thinking at the time that prettier is maybe a bit overkill.

No doubt 3 million JS linter tools will have been released since you opened this so we can reassess if needed.

Issue #295, opened 2022-04-21
Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC plugin: specified object was not found


fredrikbaberg:
Hi,
I've been trying to get Tdarr and QSV working, but am encountering an issue with the plugin: `Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC Boosh-Transcode using QSV GPU & FFMPEG`. You can see part of the output below, plugin settings are the default.

```
2022-04-21T11:51:41.091Z [hevc_qsv @ 0000019e68e37740] Could not load the requested plugin '2fca99749fdb49aeb121a5b63ef568f7': specified object was not found (-9)
2022-04-21T11:51:41.091Z [hevc_qsv @ 0000019e68e37740] Error loading plugins
2022-04-21T11:51:41.091Z 
2022-04-21T11:51:41.091Z Error initializing output stream 0:0 -- Error while opening encoder for output stream #0:0 - maybe incorrect parameters such as bit_rate, rate, width or height
2022-04-21T11:51:41.091Z 
2022-04-21T11:51:41.091Z Conversion failed!
```

I've also tried with `Tdarr_Plugin_ER01_Transcode audio and video with HW (PC and Mac)
Transcode Using QSV or VT & FFMPEG`, that works as I can see GPU being used by ffmpeg, but it touches subtitles and audio which I don't want.

Node runs on Windows 11, I just downloaded Tdarr_Updater, launched that and ran Tdarr_Node. I've setup the node to only use qsv, did not enable "Allow GPU workers to do CPU tasks", and have it set to 0 CPU, 1 GPU.

supersnellehenk:
Are you running a multi-gpu setup? You might have to plug in a monitor to the output for the intel GPU and do a reboot for it to recognize. QSV can be quite tricky. Make sure you've also got the drivers installed.

fredrikbaberg:
No multi-gpu setup.

My setup currently consists of two parts:
- Server (VM in Proxmox) with Tdarr_Server and Tdarr_Node. No access to GPU. Allocated 1 Transcode CPU, 1 Health Check CPU.
- Laptop (Windows 11) with Tdarr_Node. Just an Intel Core i3-8145U with Intel UHD Graphics 620. Allocated 1 Transcode GPU.

Drivers? The other plugin seems to be able to use GPU. I've also successfully tested Handbrake with hardware QSV on the laptop.

[...]

PR #306, opened 2022-05-08
Update plugin - Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC - OS support, custom ffmpeg cmd input + clean-up

Boosh1:
General improvements to the plugin

- Tooltips heavily cleaned up and organised so readability is better
- Added new input to so encode cmd can be customised with extra options
- Added OS support to modify the cmd based on that (Should be an actual fix for #295 )
- Reorganised the decode & encode steps so it's clearly separated
- Cleaned up decode options to only list valid HW decodes
- added "-fflags +genpts" to fix pesky avi files or anything missing timestamps


supersnellehenk (approved):

HaveAGitGat:
Looks good :)

HaveAGitGat (approved):

PR #327, opened 2022-08-25
Create Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js

KeesTucker:
Made a plugin a while back and decided to throw it up here for others to use :)
Description: 
This plugin used nvenc and transcodes based on specificed CQ:V value.
Will transcode if bitrate is greater than "optimized bitrate". Optimal bitrate accounts for fps and resolution.
Optimized bitrate can be configured using targetCodecCompression. Smaller values target lower bitrates.
FFmpeg Preset can be configured, uses slow by default.
Low res files (720p or below) will not be transcoded
If files are not in hevc they will be transcoded.
The output container is mkv.
Note that you will get an "infinite transcode loop" error if target codec compression is too low compared to CQ:V or CQ:V is too low compared to compression.
Basically: If you increment target codec compression you can lower cqv for higher quality and vice versa.
Thanks to JB and vdka for their plugins, this uses code from their work.


Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
+/* eslint-disable */
```
supersnellehenk, line 1: Remove the eslint disable. Plugins need to be checked by eslint.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+        },
+        tooltip: `This effects the target bitrate by assuming a compression ratio.
+
+      \\nExample:\\n
+      0.08`
+      },
+      {
+        name: 'cqv',
+        type: 'string',
+        defaultValue: '28',
```
supersnellehenk, line 38: Can be number

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+        },
+        tooltip: `Enter the CQ:V value, lower is higher quality.
+
+      \\nExample:\\n
+      28`
+      },
+      {
+        name: 'bframe',
+        type: 'string',
+        defaultValue: '0',
```
supersnellehenk, line 50: Can be number

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Specify amount of b-frames to use, 0-5. Use 0 to disable. (GPU must support this, turing and newer supports this, except for the 1650)
```
supersnellehenk, line 54: Earlier 1650's don't support it, from a certain SKU they put the Turing chip in there. But this is nitpicky :)

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+  const streamFPS = file.mediaInfo.track[0].FrameRate;
+  let streamBR = file.bit_rate;
+
+  //check if the file is a video, if not the function will be stopped immediately
+  if (file.fileMedium !== 'video') {
+    response.processFile = false
+    response.infoLog += '☒File is not a video! \n'
+    return response
+  } else {
+    // bitrateprobe = file.ffProbeData.streams[0].bit_rate;
```
supersnellehenk, line 116: Remove comment

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+
+  // Check if preset is configured, default to slow if not
+  var ffmpeg_preset
+  if (inputs.ffmpeg_preset === undefined) {
+    ffmpeg_preset = `slow`
+    response.infoLog += '☑Preset not set, defaulting to slow\n'
+  } else {
+    ffmpeg_preset = `${inputs.ffmpeg_preset}`
+    response.infoLog += `☑Preset set as ${inputs.ffmpeg_preset}\n`
+  }
```
supersnellehenk, line 144: This will never be undefined. Better to do `inputs.ffmpeg_preset` since that will also check if it's empty.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+    response.preset = `-c:v vc1_cuvid`
+  } else if (file.video_codec_name == 'vp8') {
+    response.preset = `-c:v vp8_cuvid`
+  } else if (file.video_codec_name == 'vp9') {
+    response.preset = `-c:v vp9_cuvid`
+  } else if (file.video_codec_name == 'hevc') {
+    if (file.ffProbeData.streams[0].profile != 'High 10') {
+      response.preset = `-c:v hevc_cuvid`
+    }
+  }
```
supersnellehenk, line 170: I don't know the exact details, but this can also be replaced with `-hwaccel cuda` instead. This makes it so the GPU/ffmpeg decides what to do.

If you decide to keep this, make it a switch statement and use strict checking. `===`

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+    response.processFile = false
+    response.infoLog += '☑File is already in hevc and is below optimal bitrate!\n'
+    return response
+  }
+  else if (file.ffProbeData.streams[0].codec_name == 'hevc'
+  && (file.video_resolution === '480p' || file.video_resolution === '576p' || file.video_resolution === '720p')) {
+    response.processFile = false
+    response.infoLog += '☑File is low res, wont transcode!\n'
+    return response
+  }
```
supersnellehenk, line 134: What happens if a file is below the optimal bitrate but is not HEVC?

You don't have to set `processFile` either, since it's already `false`.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+          file.ffProbeData.streams[i].codec_name.toLowerCase() == 'bmp' ||
+          file.ffProbeData.streams[i].codec_name.toLowerCase() == 'mjpeg') &&
+        file.ffProbeData.streams[i].codec_type.toLowerCase() == 'video'
+      ) {
+        map = `-map 0:v:0 -map 0:a -map 0:s?`
+      }
+    } catch (err) {}
+  }
+
+  response.preset += `,${map} -dn -c:v hevc_nvenc -b:v 0 -preset ${ffmpeg_preset} -cq ${inputs.cqv} -rc-lookahead 32 -bf ${inputs.bframe} -a53cc 0 -c:a copy ${subcli}${maxmux} -pix_fmt p010le`
```
supersnellehenk, line 208: Perhaps you should give the user the option if they want 10 bit or not, or at least mention it in the description.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+    try {
+      if (
+        (file.ffProbeData.streams[i].codec_name.toLowerCase() == 'png' ||
+          file.ffProbeData.streams[i].codec_name.toLowerCase() == 'bmp' ||
+          file.ffProbeData.streams[i].codec_name.toLowerCase() == 'mjpeg') &&
+        file.ffProbeData.streams[i].codec_type.toLowerCase() == 'video'
+      ) {
+        map = `-map 0:v:0 -map 0:a -map 0:s?`
+      }
+    } catch (err) {}
```
supersnellehenk, line 205: Strict `===` instead of loose `==`

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+// eslint-disable-next-line no-unused-vars
+const plugin = (file, librarySettings, inputs, otherArguments) => {
+
+    const lib = require('../methods/lib')();
+  // eslint-disable-next-line no-unused-vars,no-param-reassign
+  inputs = lib.loadDefaultValues(inputs, details);
+  var subcli = `-c:s copy`
```
supersnellehenk, line 91: Can just use quotes here instead of backticks.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+  && (file.video_resolution === '480p' || file.video_resolution === '576p' || file.video_resolution === '720p')) {
+    response.processFile = false
+    response.infoLog += '☑File is low res, wont transcode!\n'
+    return response
+  }
+
+  // Check if preset is configured, default to slow if not
+  var ffmpeg_preset
+  if (inputs.ffmpeg_preset === undefined) {
+    ffmpeg_preset = `slow`
```
supersnellehenk, line 139: Can just use quotes here instead of backticks.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,220 @@
...
+    response.infoLog += '☑Preset not set, defaulting to slow\n'
+  } else {
+    ffmpeg_preset = `${inputs.ffmpeg_preset}`
```
supersnellehenk, line 142: No need to string interpolate this, you can set it directly.

supersnellehenk (changes requested):

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
KeesTucker: It will just transcode to HEVC with the specified CQ:V value, file size will stay similar or grow slightly

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
KeesTucker: that worked! thanks

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,209 @@
...
+      \\nExample:\\n
+      3`,
+    },
+    {
+      name: 'ten_bit',
+      type: 'string',
+      defaultValue: 'true',
+      inputUI: {
+        type: 'text',
+      },
```
supersnellehenk, line 64: There's a type `boolean`, you can use the inputUI type `dropdown` with an options array with `'true'` and `'false'`. https://github.com/HaveAGitGat/Tdarr_Plugins/blob/1f6512e6b652d53db602f3e95738e69fc2a0b44d/examples/Tdarr_Plugin_pre1_Pre_Proc_Example.js#L33-L39

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,209 @@
...
+      }
+    } catch (err) {
+      response.infoLog += 'Error reading image codec!';
+    }
+  }
+
+  response.preset += `-hwaccel cuda ,${map} -dn -c:v hevc_nvenc -b:v 0 -preset ${ffmpeg_preset} -cq ${inputs.cqv}`
+    + ` -rc-lookahead 32 -bf ${inputs.bframe} -a53cc 0 -c:a copy ${subcli}${maxmux}`;
+
+  if (inputs.ten_bit === 'true') {
```
supersnellehenk, line 196: Changing the type to boolean, this can be simplified to `if (inputs.ten_bit)` :)

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,209 @@
...
+        && file.ffProbeData.streams[i].codec_type.toLowerCase() === 'video'
+      ) {
+        map = '-map 0:v:0 -map 0:a -map 0:s?';
+      }
+    } catch (err) {
+      response.infoLog += 'Error reading image codec!';
+    }
+  }
+
+  response.preset += `-hwaccel cuda ,${map} -dn -c:v hevc_nvenc -b:v 0 -preset ${ffmpeg_preset} -cq ${inputs.cqv}`
```
supersnellehenk, line 193: Replace the `,` with `<io>`, the comma is the old way of doing things.

Community/Tdarr_Plugin_Soggys_NVENC_HEVC_CQV_Optimised_Bitrate.js
```diff
@@ -0,0 +1,209 @@
...
+  const optimalVideoBitrate = Math.floor((streamHeight * streamWidth * streamFPS) * inputs.targetCodecCompression);
+  response.infoLog += `Optimal Bitrate = ${(optimalVideoBitrate / 1000000).toString()}\n`;
+  response.infoLog += `Stream Bitrate = ${(streamBR / 1000000).toString()}\n`;
+  if (file.ffProbeData.streams[0].codec_name === 'hevc'
+  && (Number.isNaN(optimalVideoBitrate) || Number.isNaN(streamBR) || streamBR < optimalVideoBitrate)) {
+    response.infoLog += '☑File is already in hevc and is below optimal bitrate!\n';
+    return response;
+  }
+  if (file.ffProbeData.streams[0].codec_name === 'hevc'
+  && (file.video_resolution === '480p' || file.video_resolution === '576p' || file.video_resolution === '720p')) {
```
supersnellehenk, line 136: You could also do this video_resolution comparison like `['480p', '576p', '720p'].includes(file.video_resolution)`. But it's just a preference thing, this works just fine.

supersnellehenk (changes requested):
Besides these minor things, looks good!

supersnellehenk (approved):

HaveAGitGat:
An issue with the plugin is that it uses `const streamBR = file.bit_rate;` so if a file has lots of audio or sub streams then that will skew the calculations. Would be best to use the video stream bitrate but that's not always available.

PR #333, opened 2022-09-15
Update Sonarr parsing && reformat

supersnellehenk:
Fix Sonarr V4 and improve filename matching to IMDB


HaveAGitGat:
@supersnellehenk should we keep this backwards compatible so you can set it to v3 if needed?

supersnellehenk:
> @supersnellehenk should we keep this backwards compatible so you can set it to v3 if needed?

Good point, please keep this open until I've tested with V3. I assumed their V3 api is in Sonarr V3 as well, but I'm not sure.

supersnellehenk:
> I can confirm the API call works in Sonarr v3 and returns the expected series.
> 
> 

Thanks for testing, haven't had the time to setup a V3 instance myself. So @HaveAGitGat if the code checks out fine it can be merged.

HaveAGitGat:
Thanks guys, will merge

PR #339, opened 2022-10-05
Create Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js

cjmanca:
Video transcode plugin (FFMPEG NVENC h265)
- Post transcode size filter, but instead of cancelling stack, will rollback to original video to allow other audio/subtitle plugins to execute. Recommended to place this as the first plugin, so the size filter will only check the video stream size difference.
- Will detect HDR and tonemap to SDR.
- Uses different FFMPEG NVENC transcoding settings for 480p,576p,720p,1080p and 4KUHD.
- 10 bit output.
- The output container is mkv. Converts or strips streams which mkv doesn't support.
- bFrame support.

### DIFF
--- a/Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
+++ b/Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
@@ -1,0 +1,631 @@
+/* eslint-disable */
+const details = () => {
+  return {
+    id: "Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback",
+    Stage: "Pre-processing",
+    Name: "Tiered FFMPEG NVENC with HDR tonemap and rollback",
+    Type: "Video",
+    Operation: "Transcode",
+    Description: `
+    - REQUIRES FFMPEG 5 or higher. Ideally just use the latest if possible. Will not work right with the version of ffmpeg included with tdar.\n
+    - Requires MediaInfo to be enabled for the library.\n
+    - Post transcode size filter, but instead of cancelling stack, will rollback to original video to allow other audio/subtitle plugins to execute.
+    Recommended to place this as the first plugin, so the size filter will only check the video stream size difference.\n
+    - Will detect HDR and tonemap to SDR.\n
+    - Uses different FFMPEG NVENC transcoding settings for 480p,576p,720p,1080p and 4KUHD.\n
+    - 10 bit output.\n
+    - The output container is mkv. Converts or strips streams which mkv doesn't support.\n
+    - bFrame support.\n
+    - Portions borrowed from Migz remux and iiDrakeii.\n\n`,
+    Version: "1.0",
+    Tags: "pre-processing,ffmpeg,video only,nvenc h265",
+    Inputs:[],
+    Inputs: [
+      {
+        name: 'upperBound',
+        type: 'number',
+        defaultValue: 90,
+        inputUI: {
+          type: 'text',
+        },
+        tooltip:
+          `Enter the upper bound % size for the new file. For example, if '90' is entered, 
+          then if the new file size is larger than 90% of the original size then the original video stream will be used instead of the transcoded video stream.`,
+      },
+      {
+        name: 'hevcBitrateCutoff',
+        type: 'string',
+        defaultValue: '2000',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the bitrate cutoff for h265 reencoding in kilobytes. 
+       \\nExample:\\n 
+      
+      2000`
+      },
+      {
+        name: 'h264BitrateCutoff',
+        type: 'string',
+        defaultValue: '1000',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the bitrate cutoff for h264 reencoding in kilobytes. 
+       \\nExample:\\n 
+      
+      1000`
+      },
+      {
+        name: 'sdCQV',
+        type: 'string',
+        defaultValue: '21',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the CQ:V value you want for 480p and 576p content. 
+       \\nExample:\\n 
+      
+      21`
+      },
+      {
+        name: 'sdMaxBitrate',
+        type: 'string',
+        defaultValue: '0',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the maximum bitrate cutoff you want for 480p and 576p content in kilobytes. Use 0 for no bitrate limit (purely based on the CQ setting).
+       \\nExample:\\n 
+      
+      3000`
+      },
+      {
+        name: 'hdCQV',
+        type: 'string',
+        defaultValue: '23',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the CQ:V value you want for 720p content.  
+      
+      \\nExample:\\n
+      23`
+      },
+      {
+        name: 'hdMaxBitrate',
+        type: 'string',
+        defaultValue: '0',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the maximum bitrate cutoff you want for 720p content in kilobytes. Use 0 for no bitrate limit (purely based on the CQ setting).
+       \\nExample:\\n 
+      
+      4000`
+      },
+      {
+        name: 'fullhdCQV',
+        type: 'string',
+        defaultValue: '25',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the CQ:V value you want for 1080p content.  
+      
+      \\nExample:\\n
+      25`
+      },
+      {
+        name: 'fullhdMaxBitrate',
+        type: 'string',
+        defaultValue: '0',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the maximum bitrate cutoff you want for 1080p content in kilobytes. Use 0 for no bitrate limit (purely based on the CQ setting).
+       \\nExample:\\n 
+      
+      5000`
+      },
+      {
+        name: 'uhdCQV',
+        type: 'string',
+        defaultValue: '27',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the CQ:V value you want for 4K/UHD/2160p content.  
+      
+      \\nExample:\\n
+      27`
+      },
+      {
+        name: 'uhdMaxBitrate',
+        type: 'string',
+        defaultValue: '0',
+        inputUI: {
+          type: 'text',
+        },
+        tooltip: `Enter the maximum bitrate cutoff you want for 4K/UHD/2160p content in kilobytes. Use 0 for no bitrate limit (purely based on the CQ setting).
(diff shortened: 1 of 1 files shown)


supersnellehenk (changes requested):
New files need tests written and need to be linted by EsLint.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
+const details = () => ({
+  id: 'Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback',
+  Stage: 'Pre-processing',
+  Name: 'Tiered FFMPEG NVENC with HDR tonemap and rollback',
+  Type: 'Video',
+  Operation: 'Transcode',
+  Description: `
+    - REQUIRES FFMPEG 5 or higher. Ideally just use the latest if possible. 
+    Will not work right with the version of ffmpeg included with tdar.\n
```
supersnellehenk, line 9: tdarr*

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+      defaultValue: '5',
+      inputUI: {
+        type: 'text',
+      },
+      tooltip: `Specify amount of b-frames to use, 0-5. Use 0 to disable. 
+      (GPU must support this, turing and newer supports this, except for the 1650)  
+      
+      \\nExample:\\n
+      5`,
+    },
```
supersnellehenk, line 172: These are all numbers, not strings :)

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+  for (let i = 1; i < file.mediaInfo.track.length; i += 1) {
+    const currTrack = file.mediaInfo.track[i];
+
+    try {
+      if (currTrack.StreamOrder !== undefined && parseInt(currTrack.StreamOrder, 10) === videoTrackIdx) {
+        videoTrack = currTrack;
+      }
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.log(err);
```
supersnellehenk, line 280: Prefer printing errors to the infoLog, since console.logs don't get captured in the job report.

Only going to comment this once, but goes for the whole file.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+
+  response.infoLog += '☑File is a video! \n';
+
+  if (file.mediaInfo.track[0].extra !== undefined
+    && file.mediaInfo.track[0].extra.ADVOVTRANSCODEDONE !== undefined
+    && file.mediaInfo.track[0].extra.ADVOVTRANSCODEDONE === 'true') {
+    response.processFile = false;
+    response.infoLog += '☑File has already been transcoded! \n';
+    return response;
+  }
```
supersnellehenk, line 333: Why does it need two checks like this? The one on line 222 does *roughly* the same.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+    if (file.video_resolution === '1080p' && (bitrateprobe / 1024) < inputs.h264BitrateCutoff) {
+      response.processFile = false;
+      response.infoLog += '☑File is below cutoff already! \n';
+      return response;
+    }
+    if (file.video_resolution === '4KUHD' && (bitrateprobe / 1024) < inputs.h264BitrateCutoff * 1.5) {
+      response.processFile = false;
+      response.infoLog += '☑File is below cutoff already! \n';
+      return response;
+    }
```
supersnellehenk, line 398: Messages are different, plus perhaps include the calculated bitrate? Like `☑File bitrate is lower than 4000 Kbps for 4K! \n`

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+  } else if (file.video_codec_name.toLowerCase() === 'vp8') {
+    response.preset = '-c:v vp8_cuvid';
+  } else if (file.video_codec_name.toLowerCase() === 'vp9') {
+    response.preset = '-c:v vp9_cuvid';
+  } else if (file.video_codec_name.toLowerCase() === 'av1') {
+    response.preset = '-c:v av1_cuvid';
+  } else if (file.video_codec_name.toLowerCase() === 'vc1') {
+    response.preset = '-c:v vc1_cuvid';
+  }
+
```
supersnellehenk, line 424: Can be reduced down to `-hwaccel cuda`. This allows the GPU to decide if it's capable of HW decoding, if not it falls back to software.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+  for (let i = 0; i < file.ffProbeData.streams.length; i += 1) {
+    // strip streams not supported by mkv
+    try {
+      if (
+        file.ffProbeData.streams[i].codec_name.toLowerCase() === 'eia_608'
+        || file.ffProbeData.streams[i].codec_name.toLowerCase() === 'timed_id3'
+        // || file.ffProbeData.streams[i].codec_name.toLowerCase() === 'dvd_subtitle'
```
supersnellehenk, line 431: Commented out dvd_subtitle?

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+
+  // check if the file is eligible for transcoding
+  // if true the neccessary response values will be changed
+  if (transcode === 1) {
+    if (filterCommand !== '') {
+      response.preset = ''; // hardware decoding doesn't seem to work with the HDR tonemapping filter. Not sure why
+      filterCommand = `-vf "${filterCommand}"`;
+    }
+
+    response.preset += `<io> ${map} -dn -c:v hevc_nvenc -profile:v main10 -preset p5 -tune hq -pix_fmt p010le `
```
supersnellehenk, line 527: Perhaps make the preset configurable? List of P1-P7.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+      + `${filterCommand} -rc:v vbr `
+      + `-multipass 2 -bufsize 600M -b:v 0 -maxrate:v ${maxBitrate} -qmin 0 -qmax ${cqv} -cq:v ${cqv} `
+      + `-rc-lookahead 32 -nonref_p 1 -a53cc 0 -threads 0 ${extraOptions} ${subcli} ${maxmux} `
+      + '-metadata ADVOVTRANSCODEDONE=true';
```
supersnellehenk, line 531: Might want to add aq-mode 3, this helps a ton in dark scenes.

supersnellehenk (changes requested):

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
cjmanca: I can rework the if on 222 and combine, but it's a different purpose. 222 is checking if the new file is larger than the old file, and reverting. This section is just seeing if the file has previously been encoded already.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
cjmanca: Messages for hevc changed. The bitrate is already being printed on lines 303-319.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
cjmanca: That does seem to work. I based this portion off of iiDrakeii's plugin. I assumed he had a good reason to do these checks.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
cjmanca: I haven't been able to confirm if it's absolutely necessary to remove those. I was running it with them commented out to see if anything errored during transcode. So far nothing has, but I'm also not sure if any of the 20k+ transcodes I've done have had a dvd_subtitle stream...

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
cjmanca: UPDATE: While it "works", it's significantly slower than the cuvid method I was using. My encodes went from about ~20 minutes to 5 hours when this was changed.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
```diff
@@ -0,0 +1,563 @@
...
+  }
+
+  // need to find a reliable test for HDR content
+  if (videoStream.color_transfer !== undefined
+    && (videoStream.color_transfer.toLowerCase() === 'smpte2084'
+    || videoStream.color_transfer.toLowerCase() === 'arib-std-b67')) {
+    isHDR = true;
+    if (filterCommand !== '') {
+      filterCommand += ',';
+    }
```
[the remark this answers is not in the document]
cjmanca: I had toyed with a couple of other filter options, which aren't currently included in the plugin. This is more just to make it easier to plug in extra filters.

Community/Tdarr_Plugin_advo_FFMPEG_NVENC_Tiered_MKV_with_rollback.js
cjmanca: I looked into this; a couple of things:
1. nvenc doesn't support aq-mode 3 (it just does nothing)
2. I tried using aq-mode 2, but the file size was exactly the same as without. I also couldn't see any visible difference.
3. I'm doing 10 bit encoding, which already does quite well in dark scenes.

PR #344, opened 2022-10-14
Remove problematic default value

buthed010203:
The tooltip explains that leaving this blank will keep all subs. That isn't the case with a non blank default value.

Also what in the world is up with this?
image


HaveAGitGat:
It's just Drdd's way of calculating the target bitrate

PR #352, opened 2022-11-07
Add files via upload

tws101:
2 Community Plugins Added

Subtitle extractor (can get any language plus the forced version of those subs as well)

HEVC transcoder using GPU NVENC... bframes and can reconvert HEVC.  Extremally configurable

### DIFF
--- a/Community/Tdarr_Plugin_Extract_SRT_Subs.js
+++ b/Community/Tdarr_Plugin_Extract_SRT_Subs.js
@@ -1,0 +1,143 @@
+// tdarrSkipTest
+const details = () => ({
+  id: 'Tdarr_Plugin_Extract_SRT_Subs',
+  Stage: 'Pre-processing',
+  Name: 'tws101_SRT_Subtitle_Extractor',
+  Type: 'Video',
+  Operation: 'Transcode',
+  Description: 'This plugin extracts embedded srt and forced srt subs in one pass inside Tdarr and will optionally remove them. \n\n '
+      + 'The subs are placed in the original folder, not the cache folder or a seperate output folder.'
+	  + 'Commentary and decriptive are skipped. The first stream of each language and the first forced stream of each language are extracted.',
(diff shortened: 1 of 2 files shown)


tws101:
Updated 3 plugins total to add...  Final number will remain 3

tws101:
Fourth  Plugin added is a ONE pass Keep on audio stream....  Handles native, undefined and foreign audio and will adjust the bit rate as well as codec and channel count.

tws101:
Added fifth plugin transcode to AV1 with CPU

tws101:
Plugin count is now 6 one had an updated name to go in line with the rest of my plugins

tws101:
Reduced to the 3 plugins that I currently maintain

Distearth:
I have been using tws101 Ultimate GPU Transcoder HDR for about a week and it's exactly what I've been looking for! Thank you. I wonder if you could edit the code a little to write the new bitrate to the transcoded file? Right now it shows the OG bitrate when it's completed.

HaveAGitGat:
@tws101 issue is that you have disabled linting and tests, make sure to remove this from top of each file ty:

```
/* eslint-disable */
// tdarrSkipTest
```

tws101:
Okay I have removed the code from each....  Thank you

For flows (Meaning my next pull request I am working on)
https://www.dropbox.com/home/Tdaar/Experimental%20Flows

Is the structure on these acceptable meaning the file and directory structure... I was not sure where to place the transcoder as it is specialty so I gave it its own folder.   Also I did an interface update should I be placing that in 1.1.0 interfaces like I did or just update 1.0.0 interfaces?    For the question here I am referencing the flows in the folder linking above I have finally got them to what looks like the standard  everyone else uses.
Also should I not submit and hold back any of these as they would be considered duplicates?

tws101:
As these plugins are many years old, and to bring them up to current eslint standards will take allot of work.. I am going to close this request until I can circle back to them in the future

PR #354, opened 2022-11-26
Two new plugins, customizable subtitle plugin and all-in-one video/audio/subtitle.

tehniemer:
The SUBS plugin will extract/copy/remove embedded text and image-based subtitles according to user defined preferences in one pass inside Tdarr. S_TEXT/WEBVTT subtitles will be removed as ffmpeg does not handle them properly.

The AIO script combines the SUBS script with video and audio conversion of files in an all-in-one solution. Added sanity checks that will flag errors for misconfigured options and invalid files.


tws101:
I am testing this now....

So far excellent, 

I am going to use it to replace my modified DRpepper extractor...

tehniemer:
> I am testing this now....
> 
> So far excellent,
> 
> I am going to use it to replace my modified DRpepper extractor...

Good to hear. I've added a few additional checks so the plugin will error instead of exiting with a transcode not required if it's misconfigured.

tws101:
I am still using the subtitle plugin...  I dont have any issues to report with it...  which one to you test out errors in?

tehniemer:
Issues seem to be resolved now and some more features added. Should be good to go now.

tws101:
SUB plugin I have been running constantly since December...

Have you considered adding also extract ASS subs   ?

tehniemer:
I could look into it, but haven't run across files that have that subtype

tehniemer:
That does sound familiar, do they have fonts that go along with the subs? I think that might be an issue. I'll see if I can find a file with them and play around a bit.

tehniemer:
@HaveAGitGat any chance of getting this one merged? I've been using them exclusively without issue since the last commit. Or are classic plugins going to be abandoned with the introduction of flows?

HaveAGitGat:
@tehniemer  thanks ideally we should not be using `// tdarrSkipTest` and should add some basic tests so various inputs are checked. So would be case of creating files with the same name here:

https://github.com/HaveAGitGat/Tdarr_Plugins/tree/master/tests/Community

Then copy content of one of the other test files and modify input/output until `npm run test` works fine

tehniemer:
@HaveAGitGat understood, but being new to the tests I'm unsure of what they are actually testing for, and I don't see any documentation about setting them up. Do you have any pointers?

HaveAGitGat:
@tehniemer just checking what outputs happen when you enter certain inputs. The first test is typically with the input left blank so that the default inputs are used and then add as many other tests as needed to check the other inputs are working correctly.

E.g. simple remux plugin:
https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/tests/Community/Tdarr_Plugin_MC93_Migz1Remux.js

Inputs in first test:
```
inputs: {},
```

Inputs in second test:
```
      inputs: {
        container: 'mp4',
        force_conform: 'true',
      },
```

tehniemer:
@HaveAGitGat Thanks for the explanation, I think I'm catching on now. In short, I need to manually build the expected output for a given input set based on the data from the file(s) located in `sampleData\media` . Is that correct?

HaveAGitGat:
@tehniemer yeah that's right, you don't need to use all the sample files. For example one is h264 and the other is h265, another has many audio streams with different languages and channel counts (good for checking audio related plugins) so just case of using the relevant ones. These plugins do a bit of everything so could use each one. In some cases a little hack like this can be used to modify the sample file data so you can test the plugin in a certain way:
https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/tests/Community/Tdarr_Plugin_MC93_MigzImageRemoval.js#L25

PR #359, opened 2022-12-28
Update for bsh1_Boosh_FFMPEG_QSV_HEVC - Fix formatting, mkvpropedit support & multi OS support

Boosh1:
Bunch of changes to help support different setups. Tested on Unraid Docker & Windows node (Win 10).

- Re-ordered options for better flow (Container & options first, then encoder settings, then bitrate limits)
- Cleaned up & formatted comments & tooltips for even more clarity
- Formatted some Log messages so they appear better
- Added mkvpropedit function to get accurate bitrate info (for MKV files). Falls back to previous bitrate calculation if mkvpropedit isn't present or fails
- Added OS switch statements so we can split behaviour based on OS. Supports Windows & Linux (Unraid).
- Mac support using VideoToolbox is enabled. This isn't QSV but allows Mac users to use the plugin if desired (Tooltip should make this clear that Mac is not QSV)


Boosh1:
Someone might need to point me in the right direction to run the tests properly on windows so I can update the test. I've ran them inside VSCode and I don't get any meaningful output.

Boosh1:
Think I've finally fixed the tests. Let me know if more detailed tests are needed and I can look into it.

Boosh1:
I recently read that Tdarr 2.00.19 runs mkvpropedit between plugins? May make my mkvpropedit addition to the plugin redundant but in my case it's helpful to run to get updated bitrate stats during the plugin run (Ensure the new HEVC file bitrate is as expected & avoid possible loops).
On some limited testing I did it still seemed to be needed but let me know if that shouldn't be the case.

HaveAGitGat:
Do you have a job report of when it's still required to run? Search for mkvpropedit to see if it's running okay

Boosh1:
Yeah I've got one. tbh it does claim to refresh with mkvpropedit but after the first pass it'll get into a loop due to the option to re-process HEVC. I'm checking for video bitrate with Media info, and falling back to overall bitrate if needs be. If there's a better way to grab video bitrate let me know.

[H16corpf7-log.txt](https://github.com/HaveAGitGat/Tdarr_Plugins/files/10760444/H16corpf7-log.txt)

Boosh1:
Yeah done some more testing and my current method of remuxing after manually calling mkvpropedit lets me grab new stats. Any other method I've tried in the plugin seems to not correctly get the new stats & then it fails due to the ability to Loop. 

Tdarr doesn't seem to flag an error between runs, but the subworker does return null, if that's an issue at all. 
```
6

2023-02-17T21:06:54.715Z oYMPrP2YD:Node[Citrix-Box]:Worker[gOBQJd3nD]:Updating file properties using mkvpropedit

7

2023-02-17T21:06:54.716Z oYMPrP2YD:Node[Citrix-Box]:Worker[gOBQJd3nD]:Updated file properties using mkvpropedit

8

2023-02-17T21:06:54.717Z oYMPrP2YD:Node[Citrix-Box]:Worker[gOBQJd3nD]:Subworker exited null
```
[oYMPrP2YD-log.txt](https://github.com/HaveAGitGat/Tdarr_Plugins/files/10771787/oYMPrP2YD-log.txt)

Boosh1:
OK so couple more changes incoming just to try and improve video bitrate calcs (prefer MediaInfo but on fail fall back and attempt ffprobe BPS checks & finally if still failing fall back to overall bitrate) & adding some -init_hw_device cmds depending on OS.

@HaveAGitGat Anyway, just to clarify what's happening with mkvpropedit. So I do checks on video bitrate as I want to ensure we accurately gauge the video bitrate to determine the new HEVC video bitrate. Since the plugin can loop I've got safety checks in place to ensure a Loop is avoided where possible.

The main thing that happens to avoid such a Loop is that I manually run mkvpropedit and get those new stats

```
    // Are the stats out of date?
    if (datStats >= statsThres) {
      statsUptoDate = true;
      response.infoLog += '☑ File stats are upto date! - Continuing...\n';
    } else {
      response.infoLog += '☒ File stats are out of date! - Will attempt to use mkvpropedit to refresh stats\n';
      try {
        if (otherArguments.mkvpropeditPath !== '') { // Try to use mkvpropedit path if it is set
          proc.execSync(`"${otherArguments.mkvpropeditPath}" --add-track-statistics-tags "${currentFileName}"`);
        } else { // Otherwise just use standard mkvpropedit cmd
          proc.execSync(`mkvpropedit --add-track-statistics-tags "${currentFileName}"`);
        }
      } catch (err) {
        response.infoLog += '☒ Error updating file stats - Possible mkvpropedit failure or file issue - '
          + ' Ensure mkvpropedit is set correctly in the node settings & check the filename for unusual characters.\n'
          + ' Continuing but file stats will likely be inaccurate...\n';
        statsError = true;
      }
      if (statsError !== true) {
        // File now updated with new stats
        response.infoLog += 'Remuxing file to write in updated file stats! \n';
        response.preset += `-fflags +genpts <io> -map 0 -c copy -max_muxing_queue_size 9999 -map_metadata:g -1 
        -metadata JBDONEDATE=${new Date().toISOString()}`;
        response.processFile = true;
        return response;
      }
    }
```
The issue is that I'm only able to get new video bitrate stats if I remux the file first, as I do at the end of that code snippet. If I don't do that then on the next run I don't get accurate video bitrate and the plugin falls back on calculating overall file bitrate which works 99% of the time. Tested on Unraid Docker as Server & Node with win10 PC as 2nd node. Both nodes seem to have that behaviour.

I guess the thing I want to highlight here is should ```proc.execSync(`mkvpropedit --add-track-statistics-tags "${currentFileName}"`);``` actually process before the plugin carries on or perhaps a full remux is actually required to ensure a new file is made with those stats written in?

Boosh1:
Done some final changes which should fix up some warnings and includes the bitrate calc changes as mentioned previously.

HaveAGitGat:
ty.

Your branch is 2 years behind the main branch so would need to go through 52 rebase steps to get it up to date. Instead I opened a new branch/PR with your changes here:
https://github.com/HaveAGitGat/Tdarr_Plugins/pull/392

Make sure your main branch is up to date before opening a PR :), can do it easily using the sync fork button.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

Will continue on the other PR.

Boosh1:
Cheers. Github is still new to me but will ensure to keep it synced up for future :)

HaveAGitGat:
No problem :)

Issue #375, opened 2023-02-15
Filter By Modified Date


Tailslide:
I see there is a 'hold' feature but I'd instead like to filter by age from file modified date (like only convert if 30 days old or older)

I tried making a plugin but I'm not sure this is possible since the plugins are javascript and file doesn't seem to contain lastModified?

HaveAGitGat:
The `statSync` object on the `file` contains info for doing this:

```
"atimeMs":1676524670205.274
"mtimeMs":1568393120000
"ctimeMs":1650864287174.0857
"birthtimeMs":1675318716310.599
"atime":"2023-02-16T05:17:50.205Z"
"mtime":"2019-09-13T16:45:20.000Z"
"ctime":"2022-04-25T05:24:47.174Z"
"birthtime":"2023-02-02T06:18:36.311Z"
```

So using `file.statSync.mtimeMs` would probably be easiest.

Tailslide:
Thanks! That should work great.

PR #378, opened 2023-02-17
Modified date filter plugin

Tailslide:
Filters files by modified date age. Also see #375


HaveAGitGat:
Very nice, thanks for linting and tests.

PR #392, opened 2023-03-10
Add boosh changes

HaveAGitGat:


HaveAGitGat:
Ref: https://github.com/HaveAGitGat/Tdarr_Plugins/pull/359

HaveAGitGat:
@Boosh1 I've updated the tests, can you check if they're as expected ty

Boosh1:
Sorry for delay. Tests seem fine to me, but do say if we we'd prefer to expand them to cover more option variations etc. Unsure how in depth tests should be.

HaveAGitGat:
Looks fine, I normally just add a test for default inputs and then a couple or so for checking the various plugin inputs are working

PR #409, opened 2023-05-12
Add 2sus_Burn_in_Subtitles_Handbrake

seansusmilch:
Burns in subs so that they'll never have to be burned in on the fly when streaming.
Allows for selectively burning in subs in certain situations.

### Example
image
When a file with the following is passed to the plugin
Audio: jpn
Subtitles: eng

The plugin will burn in the English subtitles with the "H.264 MKV 1080p30" handbrake preset while preserving the chapter markers with the "-m" argument, and output it in a ".mkv" container

### DIFF
--- a/Community/Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake.js
+++ b/Community/Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake.js
@@ -1,0 +1,172 @@
+/* eslint-disable */
+const details = () => ({
+    id: "Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake",
+    Stage: "Pre-processing",
+    Name: "2sus Burn in Subtitles Handbrake",
+    Type: "Subtitle",
+    Operation: "Transcode",
+    Description: 'This plugin can burn in subtitles based on audio track language \n\n',
+    Version: "0.01",
+    Tags: "pre-processing,handbrake,subtitles only,configurable",
+    Inputs: [
+        {
+            name: 'handbrake_preset',
+            type: 'string',
+            defaultValue: 'Fast 1080p30',
+            inputUI: {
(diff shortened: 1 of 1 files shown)


Community/Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake.js
```diff
@@ -0,0 +1,172 @@
+/* eslint-disable */
```
supersnellehenk, line 1: Plugins need to be linted and have tests written.

Community/Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake.js
```diff
@@ -0,0 +1,172 @@
...
+
+    if (
+        inputs.handbrake_preset === undefined ||
+        inputs.audio_lang === undefined ||
+        inputs.sub_lang === undefined
+    ) {
+        response.processFile = false;
+        response.infoLog += "☒ Inputs not entered! \n";
+        return response;
+    }
```
supersnellehenk, line 100: Don't need this due to the default values

supersnellehenk (changes requested):

Community/Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake.js
seansusmilch: Removed this block

Community/Tdarr_Plugin_2sus_Burn_in_Subtitles_Handbrake.js
seansusmilch: I enabled linting and resolved all the issues it found

supersnellehenk:
Still needs test(s) :)

PR #410, opened 2023-05-14
Add Plugin Customizable AV1 Transcode

Jessomadic:
I've noticed there isn't any AV1 Transcode plugins that meet what I wanted so I made one for others in the same spot as me


Jessomadic:
Made an oops and I'm fixing

Issue #381, opened 2023-02-23
Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng Pulling Wrong Series from Sonarr


sirkelsalot:
I'm new to Tdarr and testing sporadic shows and movies before I run it against my full library. So far, I've found two shows where, in both cases, this plug-in queries Sonarr API and returns the same wrong IMDB ID. The two source filenames are as below:

Squid Game - S01E01 - Red Light, Green Light [WEBRip-2160p, x265 EAC3 Atmos]
The King's Avatar - S00E04 - The King's Avatar - For the Glory [HDTV-1080p, x264 AAC]

In both cases, it returns the same incorrect IMDB ID, tt1409055 (Dragon Ball Z Kai). I am not sure how it is coming to this. Both the shows above exist in Sonarr. As a result, it deletes the wrong audio track as they should be Korean and Chinese, respectively. Instead, the true native tracks are deleted and only english remains.

2023-02-23T11:47:01.567Z Pre-processing - Remove all langs except native and English
2023-02-23T11:47:01.567Z Couldn't grab ID from Radarr
2023-02-23T11:47:01.567Z Grabbed ID (tt1409055) from Sonarr
2023-02-23T11:47:01.567Z Original language: ja, Using code: jpn
2023-02-23T11:47:01.567Z Keeping languages: Japanese, Chinese, Chinese, English, undefined
2023-02-23T11:47:01.567Z ☒No audio tracks to be removed.

Curiously, I deleted Dragon Ball Z Kai from Sonarr and re-ran for Squid Games example and this time it returned tt1898069 (American Gods). So I have no clue what is happening.

2023-02-23T11:57:50.266Z Pre-processing - Remove all langs except native and English
2023-02-23T11:57:50.266Z Couldn't grab ID from Radarr
2023-02-23T11:57:50.266Z Grabbed ID (tt1898069) from Sonarr
2023-02-23T11:57:50.266Z Original language: en, Using code: eng
2023-02-23T11:57:50.266Z Keeping languages: English, Chinese, Chinese, undefined
2023-02-23T11:57:50.266Z ☒No audio tracks to be removed. 

Also, all examples show "No audio tracks to be removed.", but audio tracks were deleted.

Jinra:
[...]

PR #418, opened 2023-06-05
alternative logic for finding sonarr series

Jinra:
This PR fixes 2 things.

1. Original logic did not account for file paths in Sonarr data that may have trailing slashes, the new logic removes the slashes when reading Sonarr paths. This also caused issues when matching with...
2. Tdarr folder names. If the Sonarr path had a trailing slash, the split and slice of -1 would cause the sonarr folder name to be "". Additionally, the current logic assumes that each cache file would be put in a parent folder under the series name, however this is not the current behavior of Tdarr. I currently have my transcoder cache as "/temp", current logic will split this to ["", "temp"], and the -2 slice would give me "". This means that the logic will match to the first series to Sonarr that has a trailing slash in the path. This does not happen the first time around since the origin folder would have a series parent folder.

New logic is to remove any null items in the array before slicing. To address the second problem, new logic will assume the file name structure (as now indicated in the plugin description) and split via " - " and take the 0 index slice of the resulting list. This may fail for people who don't have their file names starting with the folder name, or for people who don't have " - " after their series name, but it's still better than the current logic where it'll match a completely different series and strip audio from it (this is an issue for anime that matches to an english series for example). See issue #381

### DIFF
--- a/Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
+++ b/Community/Tdarr_Plugin_henk_Keep_Native_Lang_Plus_Eng.js
@@ -11,7 +11,8 @@ const details = () => ({
      (requires TMDB api key) and English.
     'Native' languages are the ones that are listed on imdb. It does an API call to 
     Radarr, Sonarr to check if the movie/series exists and grabs the IMDB id. As a last resort it 
-    falls back to the IMDB id in the filename.`,
+    falls back to the IMDB id in the filename. Sonarr detection relies on your filenames formatted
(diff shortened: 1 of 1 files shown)


supersnellehenk:
Thanks for the PR to my plugin, though, I've opened #333 a while back to fix this. Still need to check Sonarr V3 with it.

Jinra:
@supersnellehenk thanks for the update! I didn't see any traction on the issue so I didn't know there was a fix for it already. I can confirm it's working on my v3 Sonarr

PR #430, opened 2023-08-06
Fix issue with Smoove h264 NVENC plugin.

StickieBE:
Fixes issue with `undefined` being added to the ffmpeg args.
> ffmpeg.exe undefined -i

Affected other users too, see [discord discussion here](https://discord.com/channels/623392507828371476/623799920574595092/1118188978059878482).


HaveAGitGat:
Thanks

PR #432, opened 2023-08-16
Removed file re-anyalsis in Boosh QSV plugin

ArabCoders:
Hello, 

I removed the unnecessary calls to mkvpropedt & ffmpeg remux, it was causing massive disk usage without really good reason as the calls are skipped for non-mkv files anyway. It's better to separate this concern into a standalone plugin if poeple wishes to keep the mkv stats update to date.

This brings the plugin to the same speed as `Migz-Transcode Using Nvidia GPU & FFMPEG` plugin.

### DIFF
--- a/Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
+++ b/Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
@@ -357,75 +357,7 @@ const plugin = (file, librarySettings, inputs, otherArguments) => {
   }
 
   // MKVPROPEDIT - Refresh video stats
-  const intStatsDays = 7; // Use 1 week threshold for new stats
   let statsUptoDate = false;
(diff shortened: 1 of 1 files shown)


supersnellehenk:
Removing mkvpropedit would break all bitrate related checks, as ffmpeg doesn't update those fields.

ArabCoders:
> Removing mkvpropedit would break all bitrate related checks, as ffmpeg doesn't update those fields.
> 
> Tdarr has a toggle to do this in Tdarr, though, it would break stacks that are currently in use.

You are correct that might be backward incompatible change, I didn't remove the bitrate checks which already exists on file, it simply fallback to regular bitrate check if MKV doesn't have the analysis done. which is already happening in the other plugins.

ArabCoders:
No longer relevant the built in plugin `Tdarr_Plugin_00td_action_transcode` does the job of Nvidia & QSV plugins and this pull request no longer needed.

PR #434, opened 2023-08-19
[New Plugin] Keep original file datetime.

ArabCoders:
Hi, 

This plugin keep the original file datetime using data provided in `otherArguments.originalLibraryFile` property. This plugin advantage over the old one is that it does not require http request or searching the db.

### DIFF
--- a/Community/Tdarr_Plugin_AC_Keep_Dates.js
+++ b/Community/Tdarr_Plugin_AC_Keep_Dates.js
@@ -1,0 +1,69 @@
+module.exports.dependencies = [
+    'touch',
+];
+
+// tdarrSkipTest
+const details = () => ({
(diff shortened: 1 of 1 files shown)


HaveAGitGat:
Will create a flow plugin for this instead

HaveAGitGat:
I've updated the https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Community/Tdarr_Plugin_z80t_keep_original_date.js

so it uses originalLibraryFile and also works on Windows (touch.sync doesn't work on Windows) so think will stick with that one. 

We can update that one if we need any extra stuff.

ArabCoders:
@HaveAGitGat Good to know i don't use windows so i just assumed the node API would abstract that 

Thanks anyway. =)

PR #445, opened 2023-08-29
Create Tdarr_Plugin_anym001_SetDefaultFirstAudioStream.js

anym001:
New Tdarr Plugin, which set first audio stream to default.

### DIFF
--- a/Community/Tdarr_Plugin_anym001_SetDefaultFirstAudioStream.js
+++ b/Community/Tdarr_Plugin_anym001_SetDefaultFirstAudioStream.js
@@ -1,0 +1,115 @@
+/* eslint-disable */
+const details = () => ({
+  id: "Tdarr_Plugin_anym001_SetDefaultFirstAudioStream",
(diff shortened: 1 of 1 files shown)


ToasterUwU:
Any chance we can get this updated/fixed, and then merged? This is a thing i really want/need. 

I cant use it tho, since no matter what i try to set this up as a local plugin, it fails to run properly.

ToasterUwU:
Has been a while once more. @anym001 @HaveAGitGat 
Any news, any ideas, any info?

anym001:
> Any news, any ideas, any info?

I am currently using the following version locally and it works perfectly. 
@HaveAGitGat could you please adopt this as a plugin?

[SetDefaultFirstAudioStream.txt](https://github.com/user-attachments/files/17610144/SetDefaultFirstAudioStream.txt)

ToasterUwU:
Thank you, i just tried it with the thing you attached and its working great, redoing my entire Library with it now

PR #468, opened 2023-09-12
Create Tdarr_Plugin_Steez_FFMPEG_CPU_GPU_HEVC_VP9_AV1.js

SteezCram:
New TDARR plugin with encoding support for HEVC, VP9 and AV1 for both CPU and GPU using FFMPEG.


HaveAGitGat:
Thanks, looks good but looking to have auto gpu hardware detection in new plugins similar to https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/Community/Tdarr_Plugin_00td_action_transcode.js so will modify if I get time.

PR #469, opened 2023-09-15
Update wording for description of upper bound of New_file_size_check

AbeKline:
The wording was unclear and ambiguous for Upper Bound. Updated to clarify while mirroring style of Lower Bound


HaveAGitGat:
Thanks 👍

PR #470, opened 2023-09-17
Use filtergraph to specify 10-bit pixel format

jsclayton:
Fixes https://github.com/HaveAGitGat/Tdarr/issues/795

After the ffmpeg 6 upgrade, the `Boosh-Transcode using QSV GPU & FFMPEG` plugin would no longer be able to transcode to 10-bit HEVC, and would give the following error:

> Impossible to convert between the formats supported by the filter 'Parsed_null_0' and the filter 'auto_scale_0'

The reason I was given is:

> `-pix_fmt`, `-s`, and the like are just sugar for `-vf scale=[...]`, and the scale filter only operates on software frames. if you're using hardware frames, you need to insert the filter yourself, and pick the appropriate one for your use-case

This PR addresses the issue by replacing `-pix_fmt` with `-vf scale=[...]`. With this change I was able to upgrade to the latest version is Tdarr and still output 10-bit HEVC.


HaveAGitGat:
Thanks 👍

PR #493, opened 2023-10-04
Radarr and Sonarr plugins (refresh/rename files) and "Remove streams by specified property" tweak

aunefyren:
Hey, I have written four new plugins and tweaked an existing one. 

**Remove streams by specified property**
I am not sure what the rules are with editing other's plugins, but I needed it tweaked for my own purpose, so I want to share.

Because of my own actions, I had some files with corrupt subtitle streams. I wanted Tdarr to remove them based on missing stream properties. I was also worried that some edge-case could cause the plugins to remove streams I wanted to keep, so I added a filter input. 

So the changes to "Remove streams by specified property" are:

- Added a configurable input for properties. The code will check if those properties are present on streams, and remove streams where the properties are missing or null.
- Added a configurable input for codec types. This acts as a filter, meaning if other properties declare the stream to be removed, if does not matter if the codec type is wrong.

**Sonarr/Radarr**
I had annoying issues where my show/movie managers (Sonarr/Radarr) would be out of sync because Tdarr altered the files. This could range from downloading a HEVC file to replace a H264 file, even though Tdarr already converted the file to HEVC, or the managers not deleting the files they are trying to delete because they don't know they were renamed. This caused a bunch of duplicate files.

These are four plugins in total, but two main actions. Refresh a show/movie and rename an episode/movie file. To be able to perform these API calls the plugin needs to find the show/movie ID in Sonarr/Radarr, and find the episode number for renaming. To find the show/movie ID the plugin performs an API search using the folder name of the file. To find the episode number it checks the file name. 

So the requirements are a decent name of the parent folder, for example: `300 Rise of an Empire (2014)`, and the episode number in the episode file name, for example: `Game of Thrones (2011) - S05E01 - The Wars to Come [265 Bluray-1080p][AC3 5.1][x265]-MZABI`. I have been using these plugins for months now with barely any issues.

I hope these could be useful for others as well. They are currently used in my flows, and there seem to be no issues, but I will look into perhaps remaking them in the new format.


supersnellehenk:
Your refresh plugins seem overly complicated. Might want to look at mine which I want to add to Tdarr when I have the time. 

https://pastebin.com/KRtzadfT

aunefyren:
> Your refresh plugins seem overly complicated. Might want to look at mine which I want to add to Tdarr when I have the time.
> 
> https://pastebin.com/KRtzadfT

Thank you for sharing, I think you are absolutely correct. I find JavaScript formatting very confusing and I am not really sure what the best practice is around logging and errors within this project. A lot of my code is plagued by attempts to bug fix and test while developing.

Your code is much more readable, and you also managed to with both Arrs into the same plugin. 

I was not aware of the `parse` API command, but when I tried to use it on my own setup, it could not manage to find series or movie IDs using my file names. I guess our setups differ in some way, I am assuming you are using your plugin and it is functioning.

Your function does not allow both HTTP and HTTPS, but that would be a simple addition.

PR #496, opened 2023-10-05
Added `CheckVideoChannelsAmount` Flow plugin.

StickieBE:


HaveAGitGat:
Hi, thanks. Filename should be camel case. Will update when I get time.

HaveAGitGat:
Renamed and updated so gives error if no video streams

PR #502, opened 2023-10-10
Fix for HW decode & High10 - Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js

Boosh1:
Previous fix broke 10bit encodes. New fix sorts that & fixes up when HW decode should be used. H264 High10 was the problem all along, QSV seems to be happy decoding most 10bit files but High10 throws a spanner in. So now we enable SW decode for anything High10 and use normal HW decode on the rest. Pixel format cmd has a small adjustment depending on HW or SW decode.

Got some new test files to cover 10bit files better in future. Also I don't think there's test data for High10 files. Not sure if that's worth setting up now that Flows is a thing?

### DIFF
--- a/Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
+++ b/Community/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js
@@ -329,6 +329,7 @@ let extraArguments = '';
 let bitrateSettings = '';
 let inflatedCutoff = 0;
 let main10 = false;
+let high10 = false;
 let videoBR = 0;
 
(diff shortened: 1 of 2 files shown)


Boosh1:
Worth checking. Do we have enough tests for this one? I could spend a bit of time writing up more tests on a future pull

HaveAGitGat:
Ty, yeah suppose could be handy. What I typically do is this:

https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/tests/Community/Tdarr_Plugin_00td_filter_bit_depth.js#L79

So can modify the video stream in the test file to have whatever is needed for the test.

PR #510, opened 2023-10-14
Add av1 option to checkVideoCodec flow plugin

andrew-kennedy:
Given how simple this plugin already was, adding another string option should be all that's needed to check if a file is already av1.


andrew-kennedy:
@HaveAGitGat is there any way I could get this merged? Ideally we would actually allow for matching on other codecs too, such as wmv.

HaveAGitGat:
@andrew-kennedy ty, make sure to only update the ts files and then run `tsc` to produce the js files.

PR #535, opened 2023-11-11
Add support for the libdfk_aac encoder to ffmpegCommandEnsureAudioStream

Recon737:
Update the **ffmpegCommandEnsureAudioStreamAdd** plugin with the ability to transcode audio to aac using the libfdk_aac codec.


HaveAGitGat:
Thanks, we need to update the `.ts` file instead of the `.js` file. Also probably best if we add an option to enable/disable the bitrate option so it doesn't cause any unexpected consequences.

PR #540, opened 2023-11-13
Add opus support to EnsureAudioStream

andrew-kennedy:
This PR still needs to be compiled with TSC, I just made a quick edit to make it easier for @HaveAGitGat . I am doing this from github's web environment so I can't compile the typescript myself. 

For reference, ffmpeg uses `libopus` as its opus encoder so it needs the little if condition to convert this encoder name to the `opus` codec name.


HaveAGitGat:
Looks good ty

HaveAGitGat:
Didn't see you hadn't compiled the typescript, I've added a new workflow job that does this automatically when a PR is opened:
https://github.com/HaveAGitGat/Tdarr_Plugins/pull/545

So your changed should be live now

PR #570, opened 2023-12-08
2.0.0 rename plugin, regex match and replace with access to more vars.

emorris00:


emorris00:
The new plugin relies on lodash which it doesn't seem to have access to normally. On my test machine I symlinked the node_modules directory from Tdarr_Server to server/Tdarr/Plugin/FlowPlugins. So something will need to be figured out.

emorris00:
Tested your changes locally and it seems to work

emorris00:
@HaveAGitGat What still needs to be done to get this released?

HaveAGitGat:
There is an existing templating system e.g. https://docs.tdarr.io/docs/plugins/flow-plugins/basics#plugin-variable-templating

So don't need to add a new one, can use `{{{args.inputFileObj.video_resolution}}}` instead of `${args.inputFileObj.video_resolution}` unless there's something I've missed.

PR #579, opened 2023-12-28
New Plugin to clean audio and subtitle streams

cpadil:
This plugin is a combination of other plugins to clean the audio and subtitle streams. No video transcoding happens here.
    For AUDIO streams, it will look for the original language in Radarr/Sonarr or TMDB to KEEP it, 
    it will also keep any additional audio stream based on the parameter audio_language (ISO-639-2), and it will remove the rest, including commentary audio stream.
    For SUBTITLE streams, it will keep subtitles based on the parameter subtitle_language (ISO-639-2) and remove based on the parameter subtitle_codecs (e.g. hdmv_pgs_subtitle). 
    If the subtitle_language is left empty, it will remove all subtitles. 
    If a subtitle is found with a codec specified in the parameter subtitle_codecs it will be removed regardless of the parameter subtitle_language

### DIFF
--- a/Community/Tdarr_Plugin_cp92_clean_audio_subtitles.js
+++ b/Community/Tdarr_Plugin_cp92_clean_audio_subtitles.js
@@ -1,0 +1,415 @@
+/* eslint-disable no-await-in-loop */
+module.exports.dependencies = ['axios@0.27.2', '@cospired/i18n-iso-languages'];
+// tdarrSkipTest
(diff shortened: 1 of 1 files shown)


supersnellehenk:
Instead of creating a new plugin, could you add the subtitle features to my original plugin? Otherwise there will be two pieces of nearly identical code in the plugin repository.

PR #587, opened 2024-01-23
Add Flow plugin to notify Plex-Media-Reanalyzer to trigger reanalysis of media in Plex

brandon099:
I wrote a utility script/webserver (https://github.com/brandon099/plex-media-reanalyzer) that takes a POST request with the Filename or Title and sends it over to Plex to reanalyze media, and this is a plugin to use it with Tdarr. 

This plugin is very basic and currently only sends the filename as that is readily available compared to the media title Plex knows about. The existing "Send Web Request" Flow plugin does not allow for variables or parameters in the POST body, such as the filename, which is required by Plex-Media-Reanalyzer to know which media item needs to be reanalyzed.

Ultimately this allows you to do things like reorder/add streams (like 2.1 audio for compatibility) and not have to change the filename for Plex to reanalyze the file. Other options out there do not reanalyze files (like plex-autoscan), at least to my knowledge and my previous attempts at getting those working.


HaveAGitGat:
Hi, thanks. All plugins including `Send Web Request` allow using flow variables using templating:
https://docs.tdarr.io/docs/plugins/flow-plugins/basics#plugin-variable-templating

So could use `{{{args.originalLibraryFile.meta.FileName}}}` in the `Send Web Request` plugin.

Would this still be needed with that in mind?

PR #600, opened 2024-02-25
Flow Plugin notify Radarr or Sonarr, more robust and new output for file not found

### LINKED ISSUE #578: NotifyRadarrOrSonarr Flow Plugin has Filename issue
I am having failures on the flow step for Notifying Radarr or Sonarr about a file being updated. Both work properly for the majority of files, but I have run into an issue with Sonarr failing to update when the filename has certain special character in the title. For example "...!"

Marnalas:
Hi there,

I added a new version of the flow plugin notifyRadarrOrSonarr for two reasons : 
- added a more robust way to find the file and a distinct output for when the file is not found ;
- added more safeties to avoid exceptions being thrown.

Might fix #578 and fix [this issue](https://github.com/HaveAGitGat/Tdarr/issues/975). It's hard to be sure but the logs looks like what I was running into before I applied the modifications.

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/tools/notifyRadarrOrSonarr/2.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/tools/notifyRadarrOrSonarr/2.0.0/index.js
@@ -106,9 +106,9 @@ var details = function () { return ({
 exports.details = details;
 var getId = function (args, arr, arrHost, headers, fileName, refreshType) { return __awaiter(void 0, void 0, void 0, function () {
     var imdbId, id, _a, _b, _c, _d;
-    var _e, _f, _g, _h;
-    return __generator(this, function (_j) {
-        switch (_j.label) {
+    var _e, _f, _g, _h, _j;
+    return __generator(this, function (_k) {
+        switch (_k.label) {
             case 0:
                 imdbId = (_f = (_e = /\b(tt|nm|co|ev|ch|ni)\d{7,10}\b/i.exec(fileName)) === null || _e === void 0 ? void 0 : _e.at(0)) !== null && _f !== void 0 ? _f : '';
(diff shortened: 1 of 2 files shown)


Marnalas:
@HaveAGitGat  My bad about the eslint warnings and errors. I fixed it.

Marnalas:
Changing this PR to draft as I intend to modify it regarding the issue detected in this post : https://github.com/HaveAGitGat/Tdarr/issues/975
I think that the "parse" endpoint currently used in this flow plugin is a pretty inefficient way to query for the movieId. It uses the file name which, for a lot of people who follow the Plex/Jellyfin/Emby naming recommendations, is quite long and contains lots of special characters that are not friendly as an HTTP query parameter. I'm gonna look into proposing a modified version of the notify flow plugin using the "movieLookup" endpoints instead.

Marnalas:
After some rework due to the issue https://github.com/HaveAGitGat/Tdarr/issues/975, this plugin now tries to get the imdbId from the file's name + path and then use the lookup endpoints : 
- for radarr : /api/v3/movie/lookup?term=imdb:<imdbId>
- for sonarr : /api/v3/series/lookup?term=imdb:<imdbId>
If there is no imdbId found or if these endpoints failed to return the movieId/serieId, then the parse endpoint is used as a fallback.

I tested the new version thoroughly (thousands of files, series and movies) on my side and I also got confirmation from the issue's OP.

Marnalas:
@HaveAGitGat Hi, I see the checks are failing at "auto_compile_and_push" because of an error while executing a git command. I don't see more detail. Anything I can do on my side ?
(I'll ask this question on any of my PR that has this problem, feel free to answer only to one of them)

HaveAGitGat:
Thanks, looks good.

PR #602, opened 2024-02-25
Flow plugin reorder streams - fix check to determine if reorder is necessary

### LINKED ISSUE #601: [BUG] Flow plugin Reorder streams always reorder even if not needed
Hi there,

First of all, thank you for the excellent work !

I've noticed a bug with the flow plugin Reorder streams. It always reorder even if the existing order is correct. After some digging in the code, I saw this : 

1. Line 114 : the current state of the streams is stored
`const originalStreams = JSON.stringify(streams);`
2. Line 116 to 119 : all the streams are added a new property with their index
`  streams.forEach((stream, index) => {
    // eslint-disable-next-line no-param-reassign
    stream.typeIndex = index;
  });`
3. Line 219 : after the new order of the streams have determined, the new state of the streams is compared to the original one. If any of the streams' properties has changed, then the ffmpeg command is set to operate the reorder.
`if (JSON.stringify(streams) !== originalStreams)`

The problem is the step 2 always changes the streams (by adding a new property) so the comparison at step 3 always detects a change in the streams. I'll create a pull request with a solution to this problem.

Marnalas:
Hi there,

This is my proposal to fix #601.

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandRorderStreams/1.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandRorderStreams/1.0.0/index.js
@@ -80,11 +80,16 @@ var plugin = function (args) {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars,no-param-reassign
     args.inputs = lib.loadDefaultValues(args.inputs, details);
     var streams = JSON.parse(JSON.stringify(args.variables.ffmpegCommand.streams));
-    var originalStreams = JSON.stringify(streams);
     streams.forEach(function (stream, index) {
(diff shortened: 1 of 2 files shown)


HaveAGitGat:
Would it not be simpler to just move ` const originalStreams = JSON.stringify(streams);` down 4 lines rather than these other changes or am I missing something?

Marnalas:
You are perfectly right. Thank you for taking the time.

One thing to note though is comparing the whole stream array to see if some items were switched is probably overkill. That's why I made an attempt with performance in mind. That being said, not sure the trade off is worth it. I'll let you be the judge of that ;)

HaveAGitGat:
Yeah think it's fine doing it that way in this case as comparison is only being done once so the overhead isn't very much.

Added here ty:
https://github.com/HaveAGitGat/Tdarr_Plugins/pull/629

PR #604, opened 2024-03-01
Add Flow Plugin for Set Default Audio Stream by Language

ProfDoof:
Hi! I have been using Tdarr, and I need to set the default audio for different videos and so I tried to craft this plugin. I haven't been able to test it yet as I have don't have a convenient way to load this plugin at this moment and so I mostly just developed it in a github codespace. I also changed some things with regards to your getOutputStreamIndex and getOutputStreamTypeIndex logic to avoid needing to loop through the array of streams every time and then added the disposition flag. 

Feel free to adjust any part of this or let me know if there's anything you'd like me to change.


HaveAGitGat:
Will handle this in the newer PR https://github.com/HaveAGitGat/Tdarr_Plugins/pull/813 thanks

PR #607, opened 2024-03-12
Flow Plugin - RemoveStreamByProperty: extend to filter by codec_type

pavax:
I extended the FlowPlugin RemoveStreamByProperty to specify the codec-type _(audio, video, any)._ My use-case was to ensure the audio streams matching the language property are no longer included in the container as I'm going to encode them in my process. 

Without this change the plugin also removes the video stream when the language property matched. 

This change should be backwards compatible as the default is set to _any_.

**Flow Example:**
image

**Plugin Configuration:**
image


Mikec78660:
This would be greatly appreciated. I frequently run into files where the video track has a language defined or defined as und and the current remove stream by property will remove the video track when trying to remove streams that are not in the native movie language.

pavax:
I opened a new PR here:

https://github.com/HaveAGitGat/Tdarr_Plugins/pull/790

PR #612, opened 2024-03-19
Extend Check File Name Includes plugin to allow regex pattern

### LINKED ISSUE #606: [ENHANCEMENT REQUEST] Extend Check File Name Includes plugin to allow regex pattern
I have a need to select tasks based on certain patterns in the filename (e.g. checking for SXXEXX). At the moment the Check File Name Includes plugin only allows static text. Will be good if there is an option to use regex instead. Thanks!

Marnalas:
Hello there,

Resolves #606. This is my proposal for : 

1. Using regex patterns in the "Check file name includes" flow plugin
2. Having the possibility to check in the file directory too

image

image
image
image

### DIFF
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -60,5 +60,8 @@ jobs:
     - run: npm i && npm i -g typescript && rm -rdf ./FlowPlugins && tsc -v && tsc
 
     - uses: stefanzweifel/git-auto-commit-action@v5
+      if: ${{ github.event.pull_request.head.repo.full_name == 'org/repo' }}
       with:
         commit_message: Apply auto-build changes
+
+    - run: (git diff --quiet HEAD -- 2>/dev/null && echo "No uncommitted changes" || (echo "Error: Uncommitted changes found." && exit 1))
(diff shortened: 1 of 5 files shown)


Marnalas:
@HaveAGitGat Hi, I see the checks are failing at "auto_compile_and_push" because of an error while executing a git command. I don't see more detail. Anything I can do on my side ?
(I'll ask this question on any of my PR that has this problem, feel free to answer only to one of them)

HaveAGitGat:
Ok so got there in the end. As mentioned [here](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#workflow-should-run-in-base-repository) there is a workaround but can be a bit dangerous with fork PRs so I changed it so that it only auto builds PRs from the parent repo. For forks it just throws an error if someone has forgotten to commit their js changes.

Better way to do this would be to have PRs go to a `dev` branch and then I can have it auto build changes when I merge `dev` into `master`, might do that in future.

HaveAGitGat:
Aside from that looks good and works well, ty.

Marnalas:
Thank you for taking the time to describe the nature of the problem. Interesting food for thoughts !

PR #617, opened 2024-04-18
Update index.js

AznIronMan:
fix for occurrences of false 'undefined' hardwareDecoding


supersnellehenk (changes requested):
This would also require the change to be made in the typescript file, then transpiled to JavaScript

HaveAGitGat:
This is already initialized here:

https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/FlowPluginsTs/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandStart/1.0.0/index.ts#L55

I believe you got this error because you did not use the `FFmpeg Command: Start` plugin before using the other FFmpeg Command plugins.

image

PR #618, opened 2024-04-18
Update index.js

AznIronMan:
helps prevent failures with push undefined


supersnellehenk (changes requested):
This would also require the change to be made in the typescript file, then transpiled to JavaScript

HaveAGitGat:
Hi, same solution as here:

https://github.com/HaveAGitGat/Tdarr_Plugins/pull/617

PR #650, opened 2024-05-23
Patching two flow plugins 

tws101:
1.  ffmpegCommandRemoveStreamByProperty

User inputs were not being trimmed on a per element basis, leading to user error in the input field.   Each string in the input array of strings is now individually trimmed reducing user input error.  This change will lead to a better user experience with the plugin as less users will have errors.

2.  ffmpegCommand/ffmpegCommandSetContainer

Convert to MP4 was not up to standard for MP4
A. Attachments are not supported on MP4 with ffmpeg, so they must be remove on convert to MP4
B. Subs in format "ass" and "ssa" are not supported in the MP4 container so they also must be removed

Convert from odd container types was not supported
Containers with issues are 'ts', 'avi', 'mpg', 'mpeg',  When converting these 4 to an MKV or an MP4  '-fflags', '+genpts'   is required in the output argument or the conversion will fail

These changes changes will increase compatibility and reduce user errors and therefor improve user experience.

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandSetContainer/1.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandSetContainer/1.0.0/index.js
@@ -95,7 +95,7 @@ var plugin = function (args) {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


HaveAGitGat:
Looks good thanks

PR #676, opened 2024-06-21
Use yuv420p10le for libsvtav1

HaveAGitGat:


Maxmystere (approved):

Maxmystere:
After doing some tests I discovered that this change is not enough, you also need to change the profile:v to main or another supported profile since main10 doesn't exist
[Supported profiles](https://www.ffmpeg.org/ffmpeg-codecs.html#Options-29)
[Detailed profiles](https://en.wikipedia.org/wiki/AV1#Profiles)

Kitt3120:
For anyone waiting, I implemented this workaround for the time until this is fixed:

image
image

PR #680, opened 2024-06-27
allow peppershaker extract SRT plugin to process more subtitle types

fossephate:


HaveAGitGat:
Added an input for this so that it doesn't change default behaviour. LGTM thanks.

PR #688, opened 2024-07-30
Remove streamby propertyfix

rgreen83:


HaveAGitGat:
Thanks, haven't merged this as changing the variable to `valuesToCompare` will break/give different behaviour for people already using this plugin. Likewise removing the for loop logic will give different behaviour as `prop.includes(val)` is very different from `valuesToCompare.includes(prop)`. With former you can have e.g. 'mov,mp4,m4a,3gp,3g2,mj2'.includes('mov') (true) and it will work as planned but  ['mov','xxx'].includes('mov,mp4,m4a,3gp,3g2,mj2') (false) will work very differently.

rgreen83:
Updated to address issues, request review @HaveAGitGat 
New version # to avoid conflict
revert back to original naming convention per HGG
fresh start on comparison logic, tested for corner cases with invalid inputs per HGG
exclude stream 0 since this is the video stream, otherwise not_includes will remove it unless video codec specified
change match to be exact to avoid issues such as eac3 getting removed for ac3 match
change not_includes to compare against full array instead of individually to fix bug that caused everything to get removed if using multiple values
also sorry im terrible at git
image
image

Issue #583, opened 2024-01-08
QSV mode quality


Kedryn:
i just realized that the QSV (GPU) ffmpegCommandSetVideoEncoder, outputs a command with "-qp <num>", but QSV encoder uses "-global_quality <num>". That means that, when using QSV, the quality settings is ignored.

wizu:
Yes, I've now looked into my qsv process and it is running as:
`Running tdarr-ffmpeg -y -hwaccel qsv -i /media/filmy/apocalypse now/apoc.mkv -map 0:0 -c:0 hevc_qsv **-qp 19** -preset slow -profile:v:0 main10 -vf scale_qsv=format=p010le -map 0:1 -c:1 copy -map 0:3 -c:2 copy -map 0:2 -c:3 copy /temp/tdarr-workDir-node-8B8uvKG3j-worker-raw-roller-ts-1726306735899/1726306737575/apoc.mkv`

While ffmpeg wiki uses **-global_quality**
https://trac.ffmpeg.org/wiki/Hardware/QuickSync#Transcode

I'm using Flow plugin: FFmpeg Command: Set Video Encoder

Justawebguy:
I am joining in on running into this issue.

PR #694, opened 2024-09-16
Change -qp to -global_quality when qsv

wizu:
Fixes https://github.com/HaveAGitGat/Tdarr_Plugins/issues/583


HaveAGitGat:
Thanks

PR #696, opened 2024-09-21
New Flow Plugin - checkFileDuration

r0guepacket:
The new flow plugin determines the original file duration and allows a decision output if it is over a threshold number of seconds.

**Configuration Example:**
image

**Flow Example**
image


HaveAGitGat:
I suppose we could also do this using `Check Flow Variable` if >, < etc are added

r0guepacket:
> I suppose we could also do this using `Check Flow Variable` if >, < etc are added

I agree, adding further functionality to the variable check would have much greater value

PR #700, opened 2024-10-13
add ability to upmix channels with a filter

Sipnol:


FlowPluginsTs/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandEnsureAudioStream/1.0.0/index.ts
```diff
@@ -245,18 +300,27 @@ const attemptMakeStream = ({
...
+  } else {
+    streamCopy.outputArgs.push('-ac', `${targetChannels}`);
+  }
   streamCopy.outputArgs.push('-c:{outputIndex}', audioEncoder);
-  streamCopy.outputArgs.push('-ac', `${targetChannels}`);
 
   if (enableBitrate) {
-    const ffType = getFfType(streamCopy.codec_type);
-    streamCopy.outputArgs.push(`-b:${ffType}:{outputTypeIndex}`, `${bitrate}`);
+    streamCopy.outputArgs.push('-b:a:{outputTypeIndex}', `${bitrate}`);
```
Sipnol, line 313: so i removed the getType here because we filter the streams to only include audio anyway.

FlowPluginsTs/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandEnsureAudioStream/1.0.0/index.ts
```diff
@@ -245,18 +300,27 @@ const attemptMakeStream = ({
   const streamCopy: IffmpegCommandStream = JSON.parse(JSON.stringify(streamWithHighestChannel));
   streamCopy.removed = false;
   streamCopy.index = streams.length;
+  if (targetChannels > highestChannelCount && upmixingFilter !== '') {
+    streamCopy.outputArgs.push('-filter_complex', `[${streamCopy.mapArgs[1]}]${upmixingFilter}[a_{outputIndex}]`);
+    streamCopy.outputArgs.push('-map', '[a_{outputIndex}]');
+    streamCopy.mapArgs = [];
```
Sipnol, line 306: need to remove the mapArgs here because otherwise the source will be copied twice

Sipnol:
Also noticed that if you have multiple blocks of this in one ffmpeg command it will override some settings so i fixed that.

PR #723, opened 2024-11-29
add rkmpp support

koehn:
This adds support for RKMPP, included in some Rockwell ARM chips like RK3588S. Requires an ffmpeg built with RKMPP support. See https://github.com/nyanmisaka/ffmpeg-rockchip which is the ffmpeg that I’m using.


HaveAGitGat:
LTGM, thanks

PR #728, opened 2024-12-10
Create Tdarr_Plugin_DUrEX.js

tonydav:
New classic plugin

### DIFF
--- a/Community/Tdarr_Plugin_DUrEX.js
+++ b/Community/Tdarr_Plugin_DUrEX.js
@@ -1,0 +1,1000 @@
+/* eslint-disable */
+
+const details = () => {
+  return {
+    id: "Tdarr_Plugin_DUrEX",
+    Stage: 'Pre-processing',
+    Name: "Davo's Ultimate reEcoder eXtended", 
+    Type: "Video",
+    Operation: "Transcode",
+    Description: "One of if not the only plugin that will reencode HEVC to HEVC. So for example from really high bitrate bluray to quality suitable for a TV show. Will reecode video based on tiers and can optionally downscale 4k to 1080p. If target rate percentage is lower than tier will scale to that. Reencodes HEVC files but target percentage is always 1. Will remove audio and subtitles that are not in the configured language or marked as commentary. If there are multiple audio streams you can use priority options to specify your preferred stream to keep. E.g. truehd>dts>ac3. You can optionally reencode audio with bitrate per channel and desired codec and number of channels. Default is still to copy the audio though. eXtended from the original plugin to now use hw or cpu encoding and determine what to use with specific parameters for each GPU.  Finally, now will detect if previously encoded by metadata Davoprocessed=true and if so won't attempt again. This negates the need to rename the file in the flow and check for name.",
+    Version: "2.0",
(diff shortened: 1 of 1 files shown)


Community/Tdarr_Plugin_DUrEX.js
```diff
@@ -0,0 +1,1000 @@
+/* eslint-disable */
```
supersnellehenk, line 1: Please remove this and fix any ESLint errors :)

Community/Tdarr_Plugin_DUrEX.js
```diff
@@ -0,0 +1,1000 @@
...
+
+const details = () => {
+  return {
+    id: "Tdarr_Plugin_DUrEX",
```
supersnellehenk, line 5: Idk if we want a reference to *that* company in the plugin repository, but that's me.

Community/Tdarr_Plugin_DUrEX.js
```diff
@@ -0,0 +1,1000 @@
...
+        exec(command, (error) => {
+            if (jobReport) {
+                jobReport.steps.push(`Checking encoder ${encoder}: ${error ? 'Not Available' : 'Available'}`);
+            }
+            resolve(!error);
+        });
+    });
+};
+
+const getBestHevcEncoder = async (otherArguments, jobReport) => {
```
supersnellehenk, line 779: This is pretty opinionated, so it might be best to juste have a default list someone could re-order.

As in my opinion, it's CPU > QSV > NVENC > AMF, and I'm not exactly sure where Videotoolbox slots in.

dennix85: between nvenc amf
it does a pretty good job :)

100%
 CPU > QSV > NVENC > Videotoolbox > AMF,

Community/Tdarr_Plugin_DUrEX.js
tonydav: LOL. Hadn't actually considered that in the name.

Community/Tdarr_Plugin_DUrEX.js
tonydav: Excellent point. This is my personal preference as my CPU is pretty average but others have different priorities.

I'll add it in with defaults and user options as I've done with the audio selections (another personal preference thing)

tonydav: I have modified the plugin to have definable gpu/cpu priority. At this point changes via code in lines 862-869. Also changed to suggested priority order (although not personally what I use but easily modifiable if user copies to local).

	const priorities = {
		hevc_nvenc: 3,   // Lowest number - highest priority
		libx265: 1,      // don't duplicate numbers
		hevc_qsv: 2,     // numbers 1 - 6
		hevc_vaapi: 6,
		hevc_amf: 5,
		hevc_videotoolbox: 4
	};

Community/Tdarr_Plugin_DUrEX.js
tonydav: Removed, no errors

supersnellehenk: Could you upload the new version to github? Otherwise those changed won't get incorporated.

PR #734, opened 2024-12-23
Add file changed check plugin

flo1t:
Hi there

I added an additional output for the replaceOriginalFile plugin, because when no file is replaced, notifyRadarrOrSonarr causes a flow error (the PR mentioned in https://github.com/HaveAGitGat/Tdarr_Plugins/issues/578 -> https://github.com/HaveAGitGat/Tdarr_Plugins/pull/600 didn't solve this issue).


HaveAGitGat:
Thanks I've updated this PR to instead add a `Check File Changed` plugin so we can do the same checks in a more general way.

Issue #741, opened 2025-01-11
[Request] Add a flow plugin thats


Marnalas:

dennix85:
![image](https://github.com/user-attachments/assets/dd276c35-af19-4a67-a927-6dab4760cfbe)

there is one, 
i assume this is what you looking for since there is no further info in your post

Marnalas:
> ![image](https://private-user-images.githubusercontent.com/129799653/402273994-dd276c35-af19-4a67-a927-6dab4760cfbe.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzY2MTA2MjQsIm5iZiI6MTczNjYxMDMyNCwicGF0aCI6Ii8xMjk3OTk2NTMvNDAyMjczOTk0LWRkMjc2YzM1LWFmMTktNGE2Ny1hOTI3LTZkYWI0NzYwY2ZiZS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwMTExJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDExMVQxNTQ1MjRaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0wZWFmMjBiNWE0OWQ4NTVkNGI5NGJmMDBjMTEwZDgwZjVmYzRjNTgwMzk3M2RhMDQ5Njc1ZThjNmEzZDNhNGU5JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.flZbFkjdsi2_2NgQsHdSFJpDfTb91GZErpp1YxaxJBg)
> 
> there is one, i assume this is what you looking for since there is no further info in your post

Yeah, I was still writing the message but I hit 'enter' by mistake. I completed the first message.
I will create the pull request right away.

Pretty sure what I implemented doesn't already exist as reordering streams doesn't make a stream default.

PR #742, opened 2025-01-11
Flow plugin set radarr sonarr flow variables

Marnalas:
Hi,

I had the need for a flow plugin, based on the same logic as setFlowVariable, that would read info from Sonarr or Radarr then make it available to other plugins. See #741 for more details.
This is my shot at undertaking that work.

It is used this way : 
image
image

The execution will look something like that : 
image

### DIFF
--- a/.github/workflows/auto_build.yml
+++ b/.github/workflows/auto_build.yml
@@ -1,0 +1,21 @@
+name: Auto Build
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        node-version: [18.x]
+        os:
+          [
+            ["ubuntu-20.04"],
+          ]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Trigger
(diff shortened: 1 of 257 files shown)


Mikec78660 (approved):
This flow plugin seems to be working exactly as intended.

Marnalas:
@smailpouri

For the new versions of the plugins notifyRadarrOrSonarr (v3.0.0) or applyRadarrOrSonarrNamingPolicy (v2.0.0), the file info retrieval (the part that used only the imdb id) have been put in common via a separate setFlowVariablesFromRadarrOrSonarr plugin.
I have updated this plugin to also use the tvdb or tmdb ids if available.

If you want to try it you can : 
1. Change the plugins repo to https://github.com/Marnalas/Tdarr_Plugins/archive/master.zip
Image

2. Reboot Tdarr and the nodes for the repository change to apply

3. Use the plugins in a test flow : 
Make sure that this plugin is called early in your flow (before the notifyRadarrOrSonarr and applyRadarrOrSonarrNamingPolicy plugins)
image
And then use the V3.0.0 of the notifyRadarrOrSonarr plugin
image
If you were using a previous version, you need to delete it in your flow and drag and drop the plugin anew to use the most recent version.

The logs should look like this : 
image
image

PR #745, opened 2025-01-11
Flow plugin apply radarr or sonarr naming policy v2.0.0

Marnalas:
Hi,

I made a V2.0.0 of this plugin (I'm the author), that uses the Radarr or Sonarr variables made by the setFlowVariablesFromRadarrOrSonarr plugin (https://github.com/HaveAGitGat/Tdarr_Plugins/pull/742). See https://github.com/HaveAGitGat/Tdarr_Plugins/issues/741 for more details.

Still used the same way as the V1.0.0. It's mostly more effective and performant if you also use notifyRadarrOrSonarr and the setDefaultAudioStream/setDefaultSubtitleStream plugins. It also contains a modification to make the file paths it uses system agnostic.

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/tools/applyRadarrOrSonarrNamingPolicy/2.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/tools/applyRadarrOrSonarrNamingPolicy/2.0.0/index.js
@@ -97,103 +97,136 @@ var details = function () { return ({
             inputUI: {
                 type: 'text',
             },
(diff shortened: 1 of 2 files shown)


Shadowalker125:
I have an issue with the current naming policy not working at all. #798 Would this possibly be a fix for this?

Marnalas:
As far as I can guess from the logs, I don't think that it would help. This new version is mostly an optimization to reduce the number of calls to the Radarr/Sonarr API.

Issue #771, opened 2025-02-06
Flow Plugin Request - Audio Steam Number Check


Mikec78660:
A flow plugin that checks the number of audio streams. could be exactly like the audio steam check plugin. 

First output if 1 steam, second output if more than 1 stream and error output if zero streams. Would be nice if there could be an input variable for a number and have it go to first output if == to or less than that number and second output if greater than the number, and again error output if zero but might be too confusing.

If there is something like this already please let me know. I think some of the classic transcode plugins have a check to ensure at least one audio stream but I can't find anything like that in the flow plugins.

Thank you!

Marnalas:
Looks interesting and easy enough. I'll give it a go when I have some time.

Marnalas:
I've submitted a pull request for this plugin.

If you want to try it you can : 
1. Change the plugins repo to https://github.com/Marnalas/Tdarr_Plugins/archive/master.zip
<img width="947" alt="Image" src="https://github.com/user-attachments/assets/a28e55f9-9eab-4ed5-9ae2-f207ea2ccbc3" />

2. Reboot Tdarr for the repository change to apply

3. Use the plugin in a test flow : 

![Image](https://github.com/user-attachments/assets/afe0cde7-ce13-4cff-87bc-88470b2d07f8)

![Image](https://github.com/user-attachments/assets/6510ae08-2ea4-4dd1-bf56-8b4d27509656)

Logs should be something like this :
<img width="554" alt="Image" src="https://github.com/user-attachments/assets/49568938-a80e-4d0c-8c5c-759f159ab332" />

Mikec78660:
So far works great! Thank you.

HaveAGitGat:
Added `Check Stream Count`  

https://github.com/HaveAGitGat/Tdarr_Plugins/pull/981

Deprecated `Check Audio Streams Count` and `Check Video Streams Count`

PR #776, opened 2025-02-09
Flow plugin check audio streams count

Marnalas:
As requested in #771, I've written this flow plugin to add the possibility to check for a specific number of audio streams.

This plugin has three outputs : 
1. The number of audio streams is equal to the target
2. The number of audio streams is less than the target
3. The number of audio streams is more than the target

Hope it helps someone.

How to use it : 

Image

Image

Logs should be something like this :
Image
image
image


Mikec78660 (approved):

HaveAGitGat:
LGTM thanks. Added some tests and updated the outputs so ` 1 = less, 2 = equal, 3 = more` to match how other plugins do it.

PR #790, opened 2025-03-02
Flow Plugin - RemoveStreamByProperty: extend to filter by codec_type and fix issue with condition "not_includes"

pavax:
**Problem 1:**
I extended the FlowPlugin RemoveStreamByProperty to specify the codec-type (audio, video, any). My use-case was to ensure the audio streams matching the language property are no longer included in the container as I'm going to encode them in my process. Without this change the plugin also removes the video stream when the language property matched.

**Problem 2:**
also I fixed an issue when using the option: "not_includes". 

Imagine having the following audio tracks:
 
 Track 1: en
 Track 2: de
 Track 3: it
 
 Let's configure the plugin like this:
 
 Condition: not_includes
 Property to check: tags.language
 values to remove: de,en,ger,eng,und
 
 
The result should be that Track 3 is removed as it doesn't match any of the values we specified. Currently the behavior is different and won't work as all streams are removed. This is due to a bug where the code iterates over every single value-to-remove individually and marks the track as "to-remove" as soon as one value doesn't match the property-to-check. My changes fixes this behavior:

- For includes: remove if the  "Property to check" value includes ANY of the "values to remove"
- For not_includes: remove if the "Property to check" includes NONE of the "values to remove"

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandRemoveStreamByProperty/1.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandRemoveStreamByProperty/1.0.0/index.js
@@ -17,6 +17,22 @@ var details = function () { return ({
     icon: '',
     inputs: [
         {
+            label: 'Codec Type',
+            name: 'codecType',
+            type: 'string',
+            defaultValue: 'all',
+            inputUI: {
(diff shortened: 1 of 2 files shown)


pavax:
My PR should fix the issue reported by @Mikec78660 

https://github.com/HaveAGitGat/Tdarr_Plugins/issues/804

PR #791, opened 2025-03-02
Flow-Plugin: ffmpegCommandSetVdeoResolution: Intel Quick Sync (QSV) Error -> use vpp_qsv

pavax:
Fixes: Impossible to convert between the formats supported by the filter 'graph 0 input from stream 0:0' and the filter 'auto_scale_0'.

This error indicates that FFmpeg's scale filter is having trouble handling the input format when using Intel QSV (hevc_qsv). The issue is likely due to an incompatibility between the pixel format of the input file and what scale expects.

I also tried to use: `-vf vpp_qsv=w=1920:h=-2` but this is not supported by qsv. Thus, we need to set the height explicitely.


HaveAGitGat:
LTGM thanks.

PR #805, opened 2025-03-27
Use correct stream for calculating video bitrate

### LINKED ISSUE #801: ffmpegCommandSetVideoBitrate uses first stream instead of first video stream when detecting bitrate
When using ffmpegCommandSetVideoBitrate to lower a files video bitrate by using a percentage, the plugin selects the first stream and not the first video stream when detecting the base value to lower to set percentage.

So, for a file which has an audio stream as it's first stream, it selects the audio streams bitrate and then calculates the video streams output bitrate based on the audio streams bitrate.

To reproduce, use a video file that has an audio stream as it's first stream, and use a percentage of the input bitrate.

Attached is a json with the ffmpegCommand.
Look at the following lines:
- 19 for the first tracks bitrate (an audio stream)
- 128 for the second tracks bitrate (the video stream)
- 148 for the calculated output bitrate (50% of audio bitrate)

[ffmpegCommand.json](https://github.com/user-attachments/files/19383818/ffmpegCommand.json)

harryyoud:
Tracks in mediaInfo are not in the same order as ffmpeg's streams Use StreamIndex to find the correct object instead

Fixes #801

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandSetVideoBitrate/1.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandSetVideoBitrate/1.0.0/index.js
@@ -122,14 +122,14 @@ var plugin = function (args) {
     var fallbackBitrate = String(args.inputs.fallbackBitrate);
     var bitrate = String(args.inputs.bitrate);
     args.variables.ffmpegCommand.streams.forEach(function (stream) {
-        var _a, _b, _c, _d;
+        var _a, _b, _c;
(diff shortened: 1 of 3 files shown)


harryyoud:
This works locally, and fixes the issue in the description.

Before this fix, the plugin uses the first (1-indexed) item in `mediaInfo.tracks`, but this is not guaranteed to be in the same order as the ffmpeg tracks. Luckily the ffmpeg order is stored within the mediaInfo track object as StreamOrder, so we can use that instead.

harryyoud:
Hi @HaveAGitGat, this is now ready for review

HaveAGitGat:
Thanks

PR #811, opened 2025-04-08
Add Community Plugin: Remove Audio Tracks By Codec and Channels

jordanlambrecht:
This plugin will specifically remove streams that match both an audio channel and a codec name. It saves a lot of headache if a user has multiple audio codecs on the same channel count (i.e aac 5.1 and ac3 5.1)

**Example 1:** Removing 5.1 AAC Audio Tracks
A user wants to remove all 5.1 channel AAC audio tracks from their media files while preserving other audio tracks. They would set the plugin parameters to:

Codecs: aac
Channels: 5.1
Languages: (empty)

This helps users who want to save space by removing high-bandwidth AAC surround tracks when they only need stereo audio.

**Example 2:** Removing Multiple Codec Types with Language Filtering
A user with a multilingual library wants to remove both AC3 and EAC3 stereo tracks that are in Japanese, keeping only their preferred audio formats. They would configure:

Codecs: ac3,eac3
Channels: 2
Languages: jpn

This allows for targeted audio track cleanup based on both technical and language preferences.

### DIFF
--- a/Community/Tdarr_Plugin_jordy_Remove_Audio_By_Codec_Channels.js
+++ b/Community/Tdarr_Plugin_jordy_Remove_Audio_By_Codec_Channels.js
@@ -1,0 +1,208 @@
+const details = () => ({
+  id: 'Tdarr_Plugin_jordy_Remove_Audio_By_Codec_Channels',
+  Name: 'Remove Audio Tracks By Codec and Channels',
+  Type: 'Audio',
+  Operation: "Transcode",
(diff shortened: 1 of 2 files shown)


HaveAGitGat:
@jordanlambrecht you deleted this plugin and the test file in your other PR when adding `Filter by audio codec and channels`

https://github.com/HaveAGitGat/Tdarr_Plugins/pull/817/commits

image

I'll re-add it

PR #812, opened 2025-04-08
Flow plugin set radarr sonarr flow variables

Marnalas:
Hi,

I had the need for a flow plugin, based on the same logic as setFlowVariable, that would read info from Sonarr or Radarr then make it available to other plugins. See #741 for more details.
This is my shot at undertaking that work.

Usage preconditions:
- Radarr v4.x and Sonarr v4.x
- Filenames with: IMDB IDs / TVDB IDs / TMDB IDs
/!\ For files not using the IDs above, the less reliable parse method is used (match with filename).

It is used this way:
image
image

The execution will look something like that:
image


matavach:
This is awesome, and I used it for a plugin that I'm creating to work with Bazarr.

I did add need to add a line into the sonarr returns
```
    if (config.name === 'sonarr') {
      return {
        ...baseInfo,
        seasonNumber: data.parsedEpisodeInfo?.seasonNumber ?? 1,
        episodeNumber: data.parsedEpisodeInfo?.episodeNumbers?.[0] ?? 1,
        episodeId: data.episodes?.at(0).id ?? 1
      };
    }
  ```
  so that I could set the episode ID variable as well (which is needed for Bazarr
`        args.variables.user.ArrEpisodeId = String(fileInfo.episodeId ?? 0);`

Could you add that to your PR?

Marnalas:
@matavach Done ! You gave the code for when the filename is used for the search (thanks!). I also added this feature for when imdb or tvdb ids are used.

I tried the modification with every case possible : 
1. lookup on tvdbid
2. lookup on imdbid
3. parse on filename

Looks to work fine.

Marnalas:
#893 
Added ArrProfileLanguageCode variable when used with Radarr.
The variable is set with the language code of the language used in the quality profile set on the file. If the language used is "Any" then "und" is returned. If the language used is "Original" then ArrProfileLanguageCode = ArrOriginalLanguageCode.

PR #817, opened 2025-04-09
Add Community Plugin: Filter by audio codec and channels

jordanlambrecht:
Only allow specified audio codecs with specific channel counts to be processed.


HaveAGitGat:
fixed a logging bug, added some more validation and updated test, lgtm ty

PR #821, opened 2025-04-24
Changed Includes Search Folder

digitalassassins:
The `args.inputFileObj._id` folder structure turns into e.g. `C:/tmp/tdarr-workDir2-IPdHrVa0J/1745493626761/` on file load,  which defeats the object of the folder search function.

Concat `args.originalLibraryFile.meta.Directory` to working file name for correct folder search function


supersnellehenk:
Please change the Typescript file and recompile, as just the JS would get overwritten the next time.

Might even loan to bump the version, as it's a change in behaviour which might be incorporated in flows.

digitalassassins:
> Please change the Typescript file and recompile, as just the JS would get overwritten the next time.
> 
> Might even loan to bump the version, as it's a change in behaviour which might be incorporated in flows.

I don't think the search worked for folders, as the folder search did nothing previously, as it was searching the temporary working directory folder name, which is created on load. Changing it to search the `args.originalLibraryFile.meta.Directory` makes the plugin work as expected.

HaveAGitGat:
Updated

PR #822, opened 2025-04-24
Take into account Dolby Vision Profiles

digitalassassins:
Certain Dolby Vision profiles are not detected by `CheckHDR` Plugin; a check is needed of `stream.codec_tag_string` for specific Dolby codec tags as specified in the Dolby Specification:

https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile?language=en_US


supersnellehenk:
Same thing here, it needs to be added in the typescript file, then transpiled to JS :)

PR #866, opened 2025-09-06
Fix string concat error and add msmpeg4v3 codec choice

jmatthewyoung:
Update Tdarr_Plugin_MC93_Migz1FFMPEG.js

This fixes an issue with two things:

1) All .avi files cannot process due to a missing space at the +genpts command. The command would come in as tdarr-ffmpeg-fflags +genpts, and fail.

2) Codecs with msmpeg4v3 can now be processed with their correct arguments.


HaveAGitGat:
Looks good ty

PR #868, opened 2025-09-24
fix(customFunction): Use absolute paths for dynamic module require

redeven:


HaveAGitGat:
Still not sure of the reason for this change.

`outputFilePath` is an absolute path already as it is effectively `${args.workDir}/${new Date().getTime()/script.js` where the workDir is `${librarySettings.cache}/tdarr-workDir2-${job.jobId}`

The only way I can currently replicate this is if I incorrectly set the transcode cache to be just a `.`:

image

Which then causes error ` Error: Cannot find module './tdarr-workDir2-iS65v6l5j/1760945202382/script.js'`

So this error only appears when the cache folder is incorrectly set?

What are you using for your cache folder?

HaveAGitGat:
This does fix the issue but doesn't address the root cause so it's a question of whether we want to handle it like this.

PR #871, opened 2025-10-05
feat(flows): add arithmeticFlowVariable plugin

kalbasit:
A plugin to do simple arithmetic to a user variable.

### DIFF
--- a/FlowPlugins/CommunityFlowPlugins/tools/arithmeticFlowVariable/1.0.0/index.js
+++ b/FlowPlugins/CommunityFlowPlugins/tools/arithmeticFlowVariable/1.0.0/index.js
@@ -69,7 +69,6 @@ var plugin = function (args) {
     }
     var variablePath = variableMatch[1]; // e.g., "variables.user.existingVar"
     var pathParts = variablePath.split('.');
-    var variableName = pathParts[pathParts.length - 1]; // Last part for display purposes
     var variable = String(args.inputs.variable).trim();
     var operation = String(args.inputs.operation).trim();
     var quantity = parseFloat(String(args.inputs.quantity).trim());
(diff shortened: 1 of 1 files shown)


HaveAGitGat:
I've updated this so we can use it on any variable on the `args` object.

Using `const inputsOriginal = args.thisPlugin.inputsDB;` we can get the input from before the variable templating is applied, so can have the original string e.g. `{{{args.path.to.variable}}}` and can then use that when updating the variable.

FlowPluginsTs/CommunityFlowPlugins/tools/arithmeticFlowVariable/1.0.0/index.ts
```diff
@@ -0,0 +1,170 @@
...
+  const variableTemplate = String(inputsOriginal?.variable).trim();
+  const variableMatch = variableTemplate.match(/\{\{\{(?:args|baseArgs)\.([\w.]+)\}\}\}/);
+
+  if (!variableMatch) {
+    throw new Error(`Variable template "${variableTemplate}" is invalid. Expected format: {{{args.path.to.variable}}}`);
+  }
+
+  const variablePath = variableMatch[1]; // e.g., "variables.user.existingVar"
+  const pathParts = variablePath.split('.');
+  const variableName = pathParts[pathParts.length - 1]; // Last part for display purposes
```
kalbasit: Hmm @HaveAGitGat I'm not sure how this variableName is used since you seem to use `args.inputs.variable` below for the value.

digitalassassins:
@kalbasit Yeah i updated the HDR flow plugin to include Dolby Vision detection and @HaveAGitGat was very helpful. I just coded them in JS for quickness so i could hit the ground running.. I'm extremely busy with work at the moment, but soon as I get a few minuites to spare I'll modify the code to TypeScript and submit them into the repository, that is if @HaveAGitGat thinks they would be suitible to be included as Standard with the community plugins ?

I was trying to write a userguide on how to use the flow plugins to achieve different things, Hopefully i could get that finished so it could be included in the official documentation, so its not to taxing on the brain trying to work out what they are for.

Screenshot 2025-10-21 063312

Ive created examples like above, that one is to work out how long a video is and take snapshots at certain intervals through the video using calculations and the loops, aswell as creating a list. 

So it works out how many screenshots you want, gets the length of the video, works out what the times would be, then uses a loop to move ffmpeg to each specific time in the array and take a screenshot.. 

Useful to create preview screenshots for each video you encode, so you dont have to lauch each video to view the quality, you can just encode the video then get Tdarr to enter that loop at the end and boom.. you have a screenshots folder filled with the number of screenshots you want. Tdarr can then end the flow and move on to the next video..

These are just some of the advanced stuff you can do with loops, array lists and mathematics plugins 😁

PR #881, opened 2025-10-26
Boosh qsv - Format improvements & new options

Boosh1:
2 new options for more advanced uses
- Dolby Vision = For forcing processing of DV files. This is intended for uses where users have some sort of process setup to extract & inject DV data (see dovi_tools)
- HW decode failsafe = This is intended for uses where you encounter files which do fail during the HW decode (bad frame data). Intended for flows where you have 2 versions of the plugin, if the standard one fails then the 2nd should have this flag enabled so it will never attempt the HW decode.

Outside of that, mostly formatting improvements & video format flag changes to be more accurate (though likely no _functional_ difference)
- Ensure to define the format as QSV or nv12 depending on bit depth
- Change repeated if checks to array checks
- Regions added
- Minor improvements to HDR checks
- New examples & better suggestions for the extra options
- Updated hw accel flags to remove unneeded options

### DIFF
--- a/.github/workflows/auto_build.yml
+++ b/.github/workflows/auto_build.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - master
 
+permissions: {}
+
 jobs:
   test:
     strategy:
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: ['**']
 
+permissions:
+  contents: read
+
 jobs:
   test:
     strategy:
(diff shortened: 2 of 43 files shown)


PilaScat:
[Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js](https://github.com/user-attachments/files/24720882/Tdarr_Plugin_bsh1_Boosh_FFMPEG_QSV_HEVC.js)
I implemented the change myself, tested and working, I'm trying to add a commit here but without success, if you wanna check it yourself

Boosh1:
@HaveAGitGat Just FYI that I think this is good to go.

PR #903, opened 2025-12-28
Preserve original track name when downmix

PierreDurrr:
Append channels at end


HaveAGitGat:
Thanks. Added some improvements an updated tests.

HaveAGitGat:
FYI added a toggle for this behaviour `preserve_channel_title` in https://github.com/HaveAGitGat/Tdarr_Plugins/pull/974

PR #906, opened 2026-01-29
Allow https in Sonarr and Radarr URL

isthevao:


supersnellehenk:
I have considered doing this, but it would need checking if it starts with `http://` or `https://`, and if not, add the `http://` in front. Otherwise it would break every setup.

HaveAGitGat:
Looks good thanks

PR #907, opened 2026-02-12
Add splitArgs function for argument parsing

Shinrai:
Fixes issue when you need to use spaces in custom arguments.


HaveAGitGat:
Will handle this in earlier PR thanks https://github.com/HaveAGitGat/Tdarr_Plugins/pull/701

Shinrai:
> Will handle this in earlier PR thanks #701

The functions in PR 701 handle the basics but splitArgs is better because it parses the string left-to-right with state instead of jumping ahead to find matching quotes.

Main advantages:

- Handles quoted text as part of the current argument.
- Supports spaces, tabs, and newlines as separators.
- Handles nested/different quote types more predictably.
- Processes backslash escapes directly.
- Is easier to extend and debug.
- Avoids the fragile recursive search() logic in the first version.

In short: splitArgs behaves more like a real argument parser, while the first function is a simpler quote-extractor with edge-case bugs.

PR #908, opened 2026-02-22
Add AB_NotifyAndUnmonitorArr plugin

blakey108:
Post-transcode plugin that refreshes Sonarr or Radarr and optionally unmonitors  the item after transcode. Auto-detects TV vs movie from file path and routes to  the correct HD or 4K instance based on path tokens (2160p/UHD/4K). Supports dual  Sonarr and dual Radarr instances. No file management.

### DIFF
--- a/FlowPluginsTs/CommunityFlowPlugins/tools/notifyAndUnmonitorArr/1.0.0/index.ts
+++ b/FlowPluginsTs/CommunityFlowPlugins/tools/notifyAndUnmonitorArr/1.0.0/index.ts
@@ -1,0 +1,499 @@
+/* AB_NotifyAndUnmonitorArr — post-transcode Sonarr/Radarr notify + optional unmonitor
+   - Refresh Sonarr/Radarr after transcode.
+   - Optionally unmonitor the item.
+   - Sonarr unmonitor: PUT /api/v3/episode/monitor; on 404/405 fallback to PUT /api/v3/episode.
+   - Radarr unmonitor: GET /api/v3/movie/{id} → PUT same object with monitored:false.
+   - Accept S00 specials.
+   - Auto-detect app (TV→Sonarr, Movie→Radarr) from path.
+   - Route HD/4K to different Arr instances with HD fallback.
+   - No file move/copy/delete logic here.
+*/
+import {
+  IpluginDetails,
+  IpluginInputArgs,
+  IpluginOutputArgs,
+} from '../../../../FlowHelpers/1.0.0/interfaces/interfaces';
+
+/* -------------- small helpers -------------- */
+const PATH_SEP = /[\\/]/;
+const getFileName = (p: string): string => String(p || '').split(PATH_SEP).pop() || '';
+const getDirParts = (p: string): string[] => String(p || '').split(PATH_SEP).filter(Boolean);
+const stripTrailingSlashes = (u: string): string => String(u || '').replace(/\/+$/, '');
+const toBool = (v: unknown): boolean => {
+  if (typeof v === 'boolean') return v;
+  if (typeof v === 'string') return ['1', 'true', 'yes', 'on'].includes((v as string).toLowerCase());
+  return !!v;
+};
+
+interface ISxxEyy {
+  season: number | null;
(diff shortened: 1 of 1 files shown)


FlowPlugins/CommunityFlowPlugins/tools/notifyAndUnmonitorArr/1.0.0/index.js
```diff
@@ -0,0 +1,299 @@
...
+
+function parseSxxEyyFromPath(p) {
+  const base = getFileName(p);
+  const m = base.match(/S(\d{1,2})E(\d{1,3})/i);
+  return m ? { season: Number(m[1]), episode: Number(m[2]) } : { season: null, episode: null };
+}
+
+function looksLikeTvPath(p) {
+  const parts = getDirParts(p);
+  if (parts.some(x => /^Season\s+\d+$/i.test(x))) return true;
```
Rycochet: This could do with "series" as an alternative -

```suggestion
  if (parts.some(x => /^(?:Season|Series)\s+\d+$/i.test(x))) return true;
```

dennix85:
it would be easier and more in line with tdarr to split the plugin 2 parts one for radarr and one for sonarr 
as most user use other plugin that integrate with tdarr and the other starr apps, and those are also in two parts as most users assign a movies and a seperate Tv shows library folders to handle their media ;)

FlowPlugins/CommunityFlowPlugins/tools/notifyAndUnmonitorArr/1.0.0/index.js
```diff
@@ -0,0 +1,299 @@
...
+  tags: 'arr,sonarr,radarr,unmonitor,post',
+  isStartPlugin: false,
+  pType: '',
+  requiresVersion: '2.00.00',
+  sidebarPosition: -1,
+  icon: 'faBell',
+  inputs: [
+    // HD / 4K endpoints (Sonarr)
+    { label:'Sonarr HD Host',    name:'sonarr_hd_host',    type:'string', defaultValue:'', inputUI:{ type:'text' }, tooltip:'Base URL for HD Sonarr. No trailing slash.' },
+    { label:'Sonarr HD API Key', name:'sonarr_hd_api_key', type:'string', defaultValue:'', inputUI:{ type:'text' }, tooltip:'X-Api-Key for HD Sonarr' },
```
Rycochet: Do they need `HD` in the title?

I sort of feel that this should be the "normal" one, and the second `4k` one should be optional (so where they're used below it should fallback) - not everyone splits it out...

If you're adding these options you could almost have the pathname be an option for it instead of trying to magically recognise based on if there's "Season" in there, which would then allow you a more open-ended "Name" + "host" + "API key", so it could support arbitrary *arrs (that all use the same API calls)

supersnellehenk:
Flow plugins require their typescript variants to be checked into source as well.

It should also have tests like these: https://github.com/HaveAGitGat/Tdarr_Plugins/blob/master/tests/FlowPlugins/CommunityFlowPlugins/tools/applyRadarrOrSonarrNamingPolicy/1.0.0/index.test.ts

FlowPluginsTs/CommunityFlowPlugins/tools/notifyAndUnmonitorArr/1.0.0/index.ts
```diff
@@ -0,0 +1,499 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars,no-param-reassign
+  args.inputs = lib.loadDefaultValues(args.inputs, details);
+
+  const originalFileName = args.originalLibraryFile?._id || '';
+  const currentFileName = args.inputFileObj?._id || '';
+  const srcPath = currentFileName || originalFileName || '';
+
+  const isTv = looksLikeTvPath(srcPath);
+  const is4k = is4KPath(srcPath);
+  const unmonitorFlag = toBool(args.inputs.unmonitor_after_refresh);
```
supersnellehenk: Browser `Boolean` exists

PR #909, opened 2026-02-23
Add notifyRadarrOrSonarr 3.0.0 - optional unmonitor after refresh

blakey108:
Adds a 3.0.0 version of notifyRadarrOrSonarr with an optional "Unmonitor after refresh" switch (default off, so existing behaviour is unchanged).
When enabled:

Radarr: unmonitors the movie via GET → PUT /movie/{id}
Sonarr: finds the episode by SxxEyy and unmonitors via PUT /episode/monitor, with fallback to PUT /episode for older Sonarr versions


dermute:
+1 this would be an amazing feature and improve automatic handling so much!

FlowPluginsTs/CommunityFlowPlugins/tools/notifyRadarrOrSonarr/3.0.0/index.ts
```diff
@@ -0,0 +1,311 @@
...
+  fileName: string,
+): Promise<number> => {
+  const imdbId = /\b(tt|nm|co|ev|ch|ni)\d{7,10}?\b/i.exec(fileName)?.[0] ?? '';
+  let id = (imdbId !== '')
+    ? Number((await args.deps.axios({
+      method: 'get',
+      url: `${arrApp.host}/api/v3/${arrApp.name === 'radarr' ? 'movie' : 'series'}/lookup?term=imdb:${imdbId}`,
+      headers: arrApp.headers,
+    })).data?.[0]?.id ?? -1)
+    : -1;
```
Rycochet: Technically could add the other standards here
 
- `filename {imdb-tt1234567}.ext` is found by the regex already
- `filename {tvdb-1234}.ext` is only used for sonarr
- `filename {tmdb-1234}.ext` is used for both sonarr and radarr

Something along the lines of `/\{(imdb|tmdb|tvdb)-((?:tt|nm|co|ev|ch|ni)\d{7,10}?|\d+)\}/i` might work - although it'd need a slight change in behaviour... (Also, anyone with this level of naming is going to catch it with the filename below)

blakey108: Good shout, added it. Used your regex pretty much as-is, just gated per arr since the lookup endpoints want different prefixes (Radarr takes imdb:/tmdb:, Sonarr takes imdb:/tvdb:). A tag the arr can't look up, like {tmdb-} on Sonarr, falls through to the parse call same as before. So yeah, those files were getting caught either way, this just skips a round trip when the tag works.

I dropped the lazy ? while I was at it (the closing } anchors it anyway) and kept the old bare-id scan as a fallback for names like movie.tt1234567.mkv. Tests cover all three tags.

PR #947, opened 2026-04-07
Add AV1 support to rename based on codec plugin

niklas2233:
Added support for AV1 and made it change the h and x if its AV1


HaveAGitGat:
Did some refactoring and cleanup. LGTM thanks.

Issue #959, opened 2026-04-22
ffmpegCommandRemoveStreamByProperty Won’t remove non forced subtitles


Fish2:
Plugin: ffmpegCommandRemoveStreamByProperty 1.0.0

Config:
- codecType=subtitle
- propertyToCheck=disposition.forced
- valuesToRemove=1
- condition=not_includes

Expected:
- subtitle with disposition.forced=1 is kept
- subtitle with disposition.forced=0 is removed

Actual:
- non-forced subtitle is not removed

The plugin source appears to implement this using substring matching on String(target), so for target "0" and values ["1"], shouldRemove should evaluate true, but that is not what I am seeing in practice.

HaveAGitGat:
Thanks, updated

PR #960, opened 2026-04-22
fix(flows): allow falsy stream property values to be evaluated

Fish2:
Fix for https://github.com/HaveAGitGat/Tdarr_Plugins/issues/959


HaveAGitGat:
Sorry missed this, was updated with #961 , except skipping `''` so users can check empty string. Thanks.

PR #969, opened 2026-04-27
Update Tdarr_Plugin_rr01_drpeppershaker_extract_subs_to_SRT

barrymieny:
Extract HI/SDH and forced subtitles instead of overwriting normal subtitles


HaveAGitGat:
LGTM thanks

PR #983, opened 2026-05-11
fix: avoid scale_cuda for classic 10-bit NVENC

HaveAGitGat:
Summary
- Use software-frame NVDEC mode for classic 10-bit NVENC paths.
- Emit -pix_fmt p010le instead of scale_cuda to avoid filter graph reinit failures.
- Add tests for affected classic plugins and action-transcode NVENC path.


andrewd18:
Confirming that worked. Thank you!

HaveAGitGat:
@andrewd18 great thanks for letting me know!

PR #993, opened 2026-05-27
feat: add VC-1 to video codec check

HaveAGitGat:
Summary
- Add `vc1` to the Check Video Codec dropdown in TS source and generated JS.
- Add Jest coverage for the dropdown option and VC-1 detection.


dennix85:
Can we make it check for vc1 vc-1 and wvc1
Sometimes it gets returned as one above

PR #1033, opened 2026-09-17
Migz1FFMPEG: convert mov_text subtitles to subrip for mkv output

matuszeg:
### The problem

`Tdarr_Plugin_MC93_Migz1FFMPEG` and `Tdarr_Plugin_MC93_Migz1FFMPEG_CPU` both hardcode
`-c:s copy`. When the source is an MP4 carrying `mov_text` (tx3g) subtitles and the output
container is mkv, Matroska cannot store that codec, so ffmpeg aborts at header-write:

```
[matroska @ ...] Subtitle codec 94213 is not supported.
[out#0/matroska @ ...] Could not write header (incorrect codec parameters ?): Function not implemented
...
encoded 0 frames
Conversion failed!
```

The job fails having done no work, and retries fail identically. The only existing remedy
is `force_conform`, which fixes it by **dropping** the subtitles (`-map -0:i`).

I hit this on a library of MP4 sources; every affected file failed on every pass.

### The change

Emit `-c:s:N srt` for each `mov_text` subtitle stream, so the track is converted to subrip
and kept. `-c:s copy` is left in place for everything else.

Three things about the scoping, all deliberate:

- **Per stream, not global.** The subrip encoder cannot accept bitmap subtitles, so an
  unconditional `-c:s srt` would break files carrying PGS or VOBSUB that currently
  transcode fine. Targeting the individual stream index means no other track is affected.
- **Only when the output container is mkv — mp4 output is unchanged.** MP4 stores
  `mov_text` natively and cannot store subrip, so a library configured for mp4 produces
  exactly the same command as before this patch.
- **Skipped when `force_conform` is enabled for mkv.** That path already unmaps `mov_text`,
  and unmapping shifts the output stream indices the specifier refers to.

Versions bumped: `3.2` → `3.3` (GPU), `1.9` → `2.0` (CPU).

### Tests

Adds `tests/sampleData/media/sampleH264_4.json` — `sampleH264_1` plus a `mov_text`
subtitle stream — and three cases per plugin: mkv output converts, mp4 output does not,
`force_conform` does not.

`npm run test`, `npm run lint` and `npm run checkPlugins` all pass. The new cases were
written before the fix and confirmed failing; mutating the guard to be unconditional fails
the mp4 and `force_conform` cases, and removing the append fails the mkv case.

### One gap this does not close: Flows

The same file fails inside a Flow, and this PR does not help there — a Flow builds its
command from `ffmpegCommandSetVideoEncoder`, which has no conditional subtitle handling,
so the Migz plugins are never involved.

I am happy to send a second PR applying the same rule to
`FlowPluginsTs/CommunityFlowPlugins/ffmpegCommand/ffmpegCommandSetVideoEncoder`, but it is
your call where that belongs — it could equally be its own flow node rather than a change
to the encoder node. Let me know which you would prefer and I will do it that way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YLKmrWhQDev5rKdTyxferc


supersnellehenk:
Good patch, however, mp4 cannot contain image based subs, only `mov_text`. So the sole solution is adding a `-c:s srt` in case there's subtitles (not sure how ffmpeg reacts if there's none, believe it should be fine?) in an mp4 -> whatever else.

matuszeg:
Agreed on mp4. I think the codec check should stay though, mov_text shows up in m4v inputs too. And yes, `-c:s srt` with no subtitle streams is fine, exits 0 and writes nothing.

